PR #17, opened 2014-12-13
Runbot Agressive Janitor

bwrsandman:
This is a module whose aim is to save space and processing power

It checks the runbot build directory for old builds.
It checks by looking at running builds and comparing the with the contents of the build directory. Any extra builds are removed along with associated databases.
It will also do it's best at kill processes which are orphaned.

Additionally, it sets all builds as done when starting up: A big issue with runbot is that when restarting the server, the build objects are still marked as running when they're clearly not.

Credit to @foutoucour who helped out with the implementation, especially the tests.

### DIFF
--- a/runbot_janitor/models/runbot_repo.py
+++ b/runbot_janitor/models/runbot_repo.py
@@ -173,9 +173,13 @@ class RunbotRepo(models.Model):
         pattern_path = os.path.join(self.root(), 'build', pattern)
         log_dir = os.path.join(pattern_path, 'logs')
         if os.path.isdir(log_dir) and os.listdir(log_dir):
-            for directory in os.listdir(pattern_path):
-                if directory == 'logs':
+            for name in os.listdir(pattern_path):
+                path = os.path.join(pattern_path, name)
+                if name == 'logs':
                     continue
-                shutil.rmtree(directory)
+                if os.path.isdir(path)
+                        shutil.rmtree(path)
+                else:
+                    os.unlink(path)
         else:
             shutil.rmtree(pattern_path)
--- a/runbot_janitor/models/runbot_repo.py
+++ b/runbot_janitor/models/runbot_repo.py
@@ -177,7 +177,7 @@ class RunbotRepo(models.Model):
                 path = os.path.join(pattern_path, name)
                 if name == 'logs':
                     continue
-                if os.path.isdir(path)
+                if os.path.isdir(path):
                         shutil.rmtree(path)
                 else:
                     os.unlink(path)
--- a/runbot_janitor/models/runbot_repo.py
+++ b/runbot_janitor/models/runbot_repo.py
@@ -72,6 +72,7 @@ class RunbotRepo(models.Model):
         super(RunbotRepo, self).__init__(pool, cr)
         runbot_build = pool['runbot.build']
         ids = pool['runbot.build'].search(cr, SUPERUSER_ID, [])
+        _logger.info('marking %d builds as done', len(ids))
         runbot_build.write(cr, SUPERUSER_ID, ids, {'state': 'done'})
 
     @api.model
--- a/runbot_janitor/models/runbot_repo.py
+++ b/runbot_janitor/models/runbot_repo.py
@@ -103,15 +103,15 @@ class RunbotRepo(models.Model):
             try:
                 self.clean_up_database(pattern)
             except OSError as e:
-                _logger.error(e)
+                _logger.error('Error in database cleanup: %s', e)
             try:
                 self.clean_up_process(pattern)
             except OSError as e:
-                _logger.error(e)
+                _logger.error('Error in process cleanup: %s', e)
             try:
                 self.clean_up_filesystem(pattern)
             except OSError as e:
-                _logger.error(e)
+                _logger.error('Error in file system cleanup: %s', e)
 
     def clean_up_pids(self):
         """Kill all done pids which are still running
--- a/runbot_janitor/models/runbot_repo.py
+++ b/runbot_janitor/models/runbot_repo.py
@@ -27,6 +27,7 @@ import pwd
 import time
 import psutil
 import signal
+from datetime import datetime, timedelta
 
 from contextlib import closing
 
@@ -35,6 +36,7 @@ _logger = logging.getLogger(__name__)
 _logger.setLevel(logging.DEBUG)
 
 from openerp import models, api, SUPERUSER_ID, tools, sql_db
+from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT as DATETIME_FMT
 
 
 def exp_list_posix_user():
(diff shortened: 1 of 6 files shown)


runbot_janitor/models/runbot_repo.py
```diff
@@ -0,0 +1,144 @@
...
+
+from contextlib import closing
+
+import logging
+_logger = logging.getLogger(__name__)
+
+from openerp import models, api, SUPERUSER_ID, tools, sql_db
+
+
+def exp_list_posix_user():
```
foutoucour, line 42: doc string here please.
Why did you create this function and didn't use the exp_list() from runbot?

foutoucour:
:+1: LGTM even if exp_list_posix_user could use a doc string.

runbot_janitor/models/runbot_repo.py
bwrsandman, line 42: This definitely needs a docstring.

bwrsandman, line 42: done!

runbot_janitor/models/runbot_repo.py
```diff
@@ -0,0 +1,151 @@
...
+    """Rewrite/simplified version of openerp.service.exp_list()
+    Lists all databases owned by the current posix user instead of the db_user
+    from odoo config.
+    The reason for this is because runbot creates databases with the posix
+    user.
+    :returns list of databases owned by the posix user
```
foutoucour, line 48: s,:returns,:returns:
pycharm failed you ;)

max3903:
:+1:

runbot_janitor/models/runbot_repo.py
```diff
@@ -0,0 +1,154 @@
...
+        valid_builds = [b.dest for b in self.env['runbot.build'].search([
+            ('dest', 'in', list(build_dirs)),
+            ('state', '!=', 'done')
+        ])]
+        for pattern in build_dirs.difference(valid_builds):
+            _logger.info("Runbot Janitor Cleaning up Residue: %s" % pattern)
+            self.clean_up_database(pattern)
+            self.clean_up_process(pattern)
+            self.clean_up_filesystem(pattern)
+
```
gurneyalex, line 124: I would feel better with one (or three) try...except blocks inside the `for` loop, so that an OSError or any other exception on the first pattern does not prevent the cleanup from moving on. 

I agree that this is not supposed to happen, but there are system calls (`os.kill`, `shutil.rmtree`)in there which can raise exceptions, mainly OSErrors, and these deserves to at least be caught and properly logged.

gurneyalex:
great PR! 

I have but one request (see above).

runbot_janitor/models/runbot_repo.py
gurneyalex, line 124: I confirm this is required: on the OCA  instance I get:

```
2014-12-16 11:17:43,503 5000 ERROR openerp-oca-runbot openerp.addons.base.ir.ir_cron: Call of self.pool.get('runbot.repo').cron(cr, uid, *()) failed in Job 6
Traceback (most recent call last):
  File "/srv/openerp/instances/openerp-oca-runbot/parts/server/openerp/addons/base/ir/ir_cron.py", line 138, in _callback
    getattr(model, method_name)(cr, uid, *args)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/server/openerp/api.py", line 237, in wrapper
    return old_api(self, *args, **kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/server/openerp/api.py", line 332, in old_api
    result = method(recs, *args, **kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/runbot-addons/runbot_janitor/models/runbot_repo.py", line 79, in cron
    self.clean_up()
  File "/srv/openerp/instances/openerp-oca-runbot/parts/runbot-addons/runbot_janitor/models/runbot_repo.py", line 89, in clean_up
    self.clean_up_pids()
  File "/srv/openerp/instances/openerp-oca-runbot/parts/runbot-addons/runbot_janitor/models/runbot_repo.py", line 121, in clean_up_pids
    os.kill(build.pid, signal.SIGKILL)
OSError: [Errno 1] Operation not permitted
```

gurneyalex:
Hmm, actually I think the Janitor is being a bit too aggressive... I can't seem to get any build on the OCA runbot instance, and I suspect that the Janitor is killing them.

bwrsandman:
> I think the Janitor is being a bit too aggressive

I am thinking the same thing

runbot_janitor/models/runbot_repo.py
gurneyalex, line 124: the error above actually comes from a different place, but you get the idea :-)

bwrsandman:
The issue has been fixed by relaxing the janitor's pid cleanup. It was marking builds as done because it couldn't find the processes.

bwrsandman:
~~There's an issue of the log disappearing with the builds since they are in the build directory. It could make debugging annoying.~~

Solved

runbot_janitor/models/runbot_repo.py
bwrsandman, line 48: > s,:returns,:returns:

Muphry's law?

gurneyalex:
With the updated patch, I still have OSErrors on os.kill() calls:

```
2014-12-19 10:08:54,907 5664 ERROR openerp-oca-runbot openerp.addons.base.ir.ir_cron: Call of self.pool.get('runbot.repo').cron(cr, uid, *()) failed in Job 6
Traceback (most recent call last):
  File "/srv/openerp/instances/openerp-oca-runbot/parts/server/openerp/addons/base/ir/ir_cron.py", line 138, in _callback
    getattr(model, method_name)(cr, uid, *args)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/server/openerp/api.py", line 237, in wrapper
    return old_api(self, *args, **kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/server/openerp/api.py", line 332, in old_api
    result = method(recs, *args, **kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/runbot-addons/runbot_janitor/models/runbot_repo.py", line 80, in cron
    self.clean_up()
  File "/srv/openerp/instances/openerp-oca-runbot/parts/runbot-addons/runbot_janitor/models/runbot_repo.py", line 90, in clean_up
    self.clean_up_pids()
  File "/srv/openerp/instances/openerp-oca-runbot/parts/runbot-addons/runbot_janitor/models/runbot_repo.py", line 125, in clean_up_pids
    os.kill(build.pid, signal.SIGKILL)
OSError: [Errno 1] Operation not permitted
```

A typical cause for this is PID recycling by the Operating System. If you have
a stale PID in the database for a process which already exited, and the OS
reused the PID for a process owned by someone else, you will get such errors.

runbot_janitor/models/runbot_repo.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        for build in self.env['runbot.build'].search([
+            ('pid', 'in', psutil.pids()),
+            ('state', '=', 'done')
+        ]):
+            _logger.debug("Killing pid %s" % build.pid)
+            try:
+                os.kill(build.pid, signal.SIGKILL)
+            finally:
+                build.pid = False
```
gurneyalex, line 137: You want `try...except` here not `try...finally`, otherwise you exit the function at the first exception without processing the whole list of pids, and since the call to `clean_up_pids()` in `clean_up()` is not in a `try...except block`, the `clean_up` function itself exits prematurely as the exceptions gets propagated all the way up to the cron job.

gurneyalex:
I made a PR on this PR fixing the above issues https://github.com/savoirfairelinux/runbot-addons/pull/1

gurneyalex:
But then, when I manually ask for a rebuild of the 7.0 branch of OCA/OCB it gets janitored

runbot_janitor/models/runbot_repo.py
```diff
@@ -0,0 +1,171 @@
...
+class RunbotRepo(models.Model):
+    """Aggressively clean filesystem databases and processes."""
+    _inherit = "runbot.repo"
+
+    def __init__(self, pool, cr):
+        """On reinitialisation of db, mark all builds as done"""
+        super(RunbotRepo, self).__init__(pool, cr)
+        runbot_build = pool['runbot.build']
+        ids = pool['runbot.build'].search(cr, SUPERUSER_ID, [])
+        runbot_build.write(cr, SUPERUSER_ID, ids, {'state': 'done'})
```
gurneyalex, line 84: This line caused me much harm. When running with multiple workers, each worker will call this **init** method, and clean up the builds. But Odoo does not start all the workers at the same time (I did not go too deep in this), but the net result is: each time a worker starts, the builds are passed to done, and they get cleaned up afterwards by the janitor. 

I propose adding a check on the write_date here, and I'll update my PR on your branch accordingly

gurneyalex:
:+1:

PR #20, opened 2014-12-16
handle absence of gitlab module

gurneyalex:
don't crash the instance at startup if gitlab module is not available.

### DIFF
--- a/runbot_gitlab/runbot_repo.py
+++ b/runbot_gitlab/runbot_repo.py
@@ -23,8 +23,14 @@
 import re
 import logging
 from urllib import quote_plus
+
 import requests
-from gitlab3 import GitLab
+try:
+    from gitlab3 import GitLab
+except ImportError as exc:
+    # don't fail at load if gitlab module is not available
+    pass
+
 from openerp import models, fields, api
 from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT
 from openerp.tools.translate import _


pedrobaeza:
:+1:

bwrsandman:
The module won't install without `gitlab3` being available.
`gitlab3` is only imported in `runbot_repo`.
Wouldn't it be cleaner to put the `try except ImportError` block there

runbot_gitlab/__init__.py
```diff
@@ -20,9 +20,12 @@
...
-    runbot_build,
-    controllers,
-)
+try:
+    from . import (runbot_repo,
+                   runbot_branch,
+                   runbot_build,
+                   controllers,
+                   )
+except ImportError, exc:
```
bwrsandman: This is a deprecated way of catching errors, please use either

``` python
except ImportError:
except ImportError as exc:
```

gurneyalex:
@bwrsandman I updated the fix with your remarks.

runbot_gitlab/__init__.py
```diff
@@ -20,9 +20,8 @@
...
-    runbot_repo,
-    runbot_branch,
-    runbot_build,
-    controllers,
-)
+from . import (runbot_repo,
+               runbot_branch,
+               runbot_build,
+               controllers,
+               )
```
bwrsandman: I don't see the point of including these changes, can you uncommit this file?

gurneyalex:
@bwrsandman reverted **init**.py changes.

bwrsandman:
:+1:

PR #22, opened 2015-01-16
Only run init once when server is started

bwrsandman:
Turns out `__init__` is run more than once.

Add initialized global so cleaning of builds is done only once.


pedrobaeza:
:+1:

guewen:
:+1:

PR #24, opened 2015-01-21
Add lock file for prebuild segment

bwrsandman:
Some builds were not finishing prebuild section, I think it might be because there are no lock files.


gurneyalex:
:+1:

yvaucher:
:+1:

PR #32, opened 2015-04-01
Use runbot's new job_00_init for prebuild

bwrsandman:
Misc convert of syntax to v8


yvaucher:
:+1:

pedrobaeza:
Seems OK

:+1:

llacroix:
job_00_init is gone in new version of runbot. Calling super will not work anymore.

PR #33, opened 2015-04-20
Remove accents from MR name

bwrsandman:
Fix bug in urllib quote when building MR with branch in name


runbot_gitlab/runbot_repo.py
```diff
@@ -43,13 +44,29 @@
...
+def strip_accents(unicode_string):
+    """Remove accents and greek letters from string
+
+    :param unicode_string: String with possible accents
+    :type unicode_string: unicode
+    :return: String of unicode_string without accents
+    :rtype: unicode
+    """
+    return ''.join(
+        char for char in unicodedata.normalize('NFD', unicode_string)
```
pedrobaeza, line 64: Why not using unidecode that make the change in one call?

bwrsandman, line 64: It's to avoid having them in the url

yvaucher:
:+1:

pedrobaeza:
:+1:

PR #48, opened 2015-07-16
[ADD] runbot_pylint: Add module to run pylint checks in runbot

moylop260:


runbot_pylint/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+
+{
+    'name': 'Runbot Pylint',
+    'category': 'Website',
+    'summary': 'Runbot',
+    'version': '1.0',
+    'author': 'Vauxoo, Odoo Community Association (OCA)',
+    'depends': ['runbot'],
+    'external_dependencies': {
+        'bin': ['pylint'],
```
bwrsandman: I'd rather this be python instead of bin.
Pylint is a python dependency and can be installed via pip or system packager, either way:

``` bash
$ which pylint
/usr/bin/pylint
```

``` bash
$ pip2 list | grep pylint
pylint (1.4.4)
```

``` python
import pylint
assert pylint
```

runbot_pylint/models/runbot_build.py
```diff
@@ -0,0 +1,217 @@
...
+"""
+This module is used to create new fields in the inherited classes
+to work with pylint from runbot
+"""
+
+from openerp import models, api, fields
+import logging
+import os
+import stat
+from openerp.tools.safe_eval import safe_eval
```
bwrsandman, line 21: Small nitpick about the [recommended import order](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports)

runbot_pylint/models/runbot_build.py
```diff
@@ -0,0 +1,217 @@
...
+                addons_path = addons_path.strip()
+                fname_openerp = os.path.join(addons_path, module,
+                                             '__openerp__.py')
+                if os.path.isfile(fname_openerp):
+                    module_depends_list = safe_eval(
+                        open(fname_openerp, "r").read()).get('depends', [])
+                    if module_depends_list:
+                        module_depends_str = ','.join(module_depends_list)
+                        get_depends(module_depends_str, addons_paths,
+                                    depends=depends)
```
bwrsandman: It's strange to have a recursive method and not store the result after the call.

runbot_pylint/models/runbot_build.py
```diff
@@ -0,0 +1,217 @@
...
+            depends.append(module)
+            for addons_path in (addons_paths or '').split(','):
+                addons_path = addons_path.strip()
+                fname_openerp = os.path.join(addons_path, module,
+                                             '__openerp__.py')
+                if os.path.isfile(fname_openerp):
+                    module_depends_list = safe_eval(
+                        open(fname_openerp, "r").read()).get('depends', [])
+                    if module_depends_list:
+                        module_depends_str = ','.join(module_depends_list)
```
bwrsandman: Why not use the list?

runbot_pylint/models/runbot_build.py
```diff
@@ -0,0 +1,217 @@
...
+from openerp.tools.safe_eval import safe_eval
+
+
+_logger = logging.getLogger(__name__)
+
+
+def get_depends(modules, addons_paths, depends=None):
+    """
+    Get recursive depends from addons_paths and modules list
+    """
```
bwrsandman, line 34: Could you add sphinx type `:param`, `:type` and `:return` documentation for this function, it is a little confusing with the arguments and their types.

runbot_pylint/models/runbot_build.py
```diff
@@ -0,0 +1,217 @@
...
+import logging
+import os
+import stat
+from openerp.tools.safe_eval import safe_eval
+
+
+_logger = logging.getLogger(__name__)
+
+
+def get_depends(modules, addons_paths, depends=None):
```
bwrsandman: Seems like `modules` and `depends` could be sets. It would make more sense

runbot_pylint/models/runbot_build.py
```diff
@@ -0,0 +1,217 @@
...
+                fname_openerp = os.path.join(addons_path, module,
+                                             '__openerp__.py')
+                if os.path.isfile(fname_openerp):
+                    module_depends_list = safe_eval(
+                        open(fname_openerp, "r").read()).get('depends', [])
+                    if module_depends_list:
+                        module_depends_str = ','.join(module_depends_list)
+                        get_depends(module_depends_str, addons_paths,
+                                    depends=depends)
+    return depends
```
bwrsandman: This function can be simplified. It is very confusing. Here's a simplified version:

``` python
from itertools import ifilter, imap

def get_depends_simplified(modules, addons_paths):
    """
    Get recursive depends from addons_paths and modules list
    :param str modules: comma separated list of modules
    :param str addons_paths: comma separated list of paths for modules
    :yields list: Unsorted list of recursive dependencies of modules
    """
    modules = set(modules.split(","))
    addons_paths = set(map(str.strip, (addons_paths or '').split(',')))
    visited = set()
    while(modules != visited):
        module = (modules - visited).pop()
        visited.add(module)
        manifest_path = os.path.join(module, '__openerp__.py')
        try:
            manifest_filename = next(ifilter(
                os.path.isfile, 
                imap(lambda p: os.path.join(p, manifest_path), addons_paths)
            ))
        except StopIteration:
            # For some reason the module wasn't found
            continue
        manifest = safe_eval(open(manifest_filename).read())
        modules.update(manifest.get('depends', []))
        yield module  # To make into a generator
    # return sorted(modules)  # To make into a function that returns a sorted list
    # return modules  # To return a set
```

It also gains a little performance

```
In [189]: %timeit get_depends(module, addons_paths)
10 loops, best of 3: 27.7 ms per loop

In [190]: %timeit sorted(get_depends_simplified(module, addons_paths))
10 loops, best of 3: 22.3 ms per loop
```

And returns the same:

``` python
>>> sorted(get_depends_simplified(module, addons_paths)) == sorted(get_depends(module, addons_paths))
True
```

runbot_pylint/models/runbot_build.py
```diff
@@ -0,0 +1,217 @@
...
+
+            # get all depends and sub-depends from modules
+            _, modules = build.cmd()
+            depends = set(get_depends(modules, build.server('addons')))
+
+            # get all modules to check pylint intersection with modules depends
+            modules_to_check_pylint = list(depends & modules_to_check_pylint)
+        return modules_to_check_pylint
+
+    def job_15_pylint(self, cr, uid, build, lock_path, log_path, args=None):
```
bwrsandman, line 130: Why is this 15 and not 05?
Wouldn't we want this running before job_10_base and after job_00_init?

runbot_pylint/models/runbot_build.py
```diff
@@ -0,0 +1,217 @@
...
+        pylint_log = build.path('logs', 'job_15_pylint.txt')
+        pylint_error = False
+        max_log_lines = 20
+        count = 0
+        if os.path.isfile(pylint_log):
+            with open(pylint_log) as fpylint_log:
+                for line in fpylint_log:
+                    if not pylint_error and '****' in line:
+                        pylint_error = True
+                    if pylint_error:
```
bwrsandman: Couldn't these two if statements be combined?

runbot_pylint/models/runbot_repo.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    Added pylint_conf_path field to use a configuration
+      of pylint by repository,
+      to use for each build of repository.
+    """
+
+    _inherit = 'runbot.repo'
+
+    pylint_conf_path = fields.Char(
+        help='Relative path to pylint conf file')
```
bwrsandman: Missing string?

runbot_pylint/models/runbot_repo.py
```diff
@@ -0,0 +1,59 @@
...
+    check_pylint = fields.Boolean(
+        help='Check pylint to modules of this repo')
```
bwrsandman, line 34: Missing string?

moylop260, line 34: v8 don't need string if this one is the same that name of field.

runbot_pylint/models/runbot_build.py
moylop260: pylint_error is a flag.
Just need be activate one time.

runbot_pylint/models/runbot_build.py
moylop260: I will to use your method! :+1:

runbot_pylint/models/runbot_build.py
bwrsandman: Yes but the way this is set up is:
1. `pylint_error is False`
2.  `'****' in line is False`
3. both ifs are skipped
4. loop
5. `'****' in line is True`
6. first if is evaluated and `pylint_error is True` 
7. second if is evaluated
8. loop breaks out of loop
9. first is never again evaluated because `pylint_error is True`, it doesn't make a difference anyways because the only command is to make it `pylint_error is True` again
10. second if is always run from now on

Essentially, the first if is only there to be run once and it being run once makes the second if run every time after the first is run.

Say we simplied the first if to be:

``` python
if '****' in line:
    pylint_error = True
```

It wouldn't change a thing since `pylint_error` would either go from False to True like before or True to True. The only reason I see `not pylint_error` being useful is if `'****' in line` was a costly operation and it doesn't seem that way

bwrsandman: What I don't understand is why every line after `'****'` in the log is counted as a pylint error...

moylop260: We have lines with `****` and lines without `****`
You can see this example:
https://docs.google.com/file/d/0BwPeHBuUYqNseURZUGVwQUYtQ0U/edit

moylop260:
@bwrsandman 
I changed the detection fails method.

runbot_pylint/models/runbot_build.py
moylop260, line 130: @bwrsandman 
I saw that `job_00_init` don't exists in [8.0](https://github.com/odoo/odoo-extra/blob/8.0/runbot/runbot.py#L751)
I will to revert to 15

runbot_pylint/models/runbot_build.py
moylop260: @bwrsandman 
I have next traceback

``` python
set(map(str.strip, u'unicodestr'.split(",")))
Traceback (most recent call last):
  File "<input>", line 1, in <module>
TypeError: descriptor 'strip' requires a 'str' object but received a 'uni
code'
```

Do you have a problem if I change to:

``` python
set(map(lambda var: var.strip(), u'unicodestr'.split(",")))
```

runbot_pylint/models/runbot_build.py
bwrsandman: Could you get the number of violations directly from pylint instead of analyzing the log like that?

runbot_pylint/models/runbot_build.py
```diff
@@ -0,0 +1,243 @@
...
+        except StopIteration:
+            # For some reason the module wasn't found
+            continue
+        manifest = safe_eval(open(manifest_filename).read())
+        modules.update(manifest.get('depends', []))
+        # To make into a generator
+        # yield module
+
+    # To make into a function that returns a sorted list
+    # return sorted(modules)
```
bwrsandman: You can take these comments out, they were just examples. :P

moylop260:
@bwrsandman 
Changes applied!

runbot_pylint/models/runbot_build.py
bwrsandman, line 130: We're based on master, it's the most current code, see .travis.yml

bwrsandman:
I still think the pylint should be before the base install

moylop260:
@bwrsandman 
I'm agree with you.
But this is a module to work with v8 you don't have a `job_00_init`
https://github.com/odoo/odoo-extra/blob/8.0/runbot/runbot.py#L751-L762

In master version we will can change this order.

bwrsandman:
As stated before. This repo here uses the master version of odoo-extra. The modules should, therefore take that into accoun.t

moylop260:
If we have a branch `8.0` using a instance `master` there is a error.
The global standard is use `8.0` branch name for instances `8.0`
`7.0` branch name for instance `7.0`... and `master` branch name for instance `master`
You can to create a issue with this matter.

I will create a new PR to new branch `master` to use `master` instance.

bwrsandman:
I agree with you, but the fact is that we have no control over what odoo does with their repos.
That doesn't mean we need to limit ourselves to the older code of runbot (on 8.0) because of these standards.

moylop260:
@bwrsandman 
I don't understand your point sorry.

I saw the same standard with odoo repos:
odoo-odoo

They have [odoo-extra/master](https://github.com/odoo/odoo-extra/tree/master) and [odoo-extra/8.0](https://github.com/odoo/odoo-extra/tree/8.0)

And we have [oca/8.0](https://github.com/OCA/runbot-addons/tree/8.0) but we don't have [oca/master](https://github.com/OCA/runbot-addons/tree/master)

I'm generating changes in [oca/8.0](https://github.com/OCA/runbot-addons/tree/8.0) then we will need to use  [odoo-extra/8.0](https://github.com/odoo/odoo-extra/tree/8.0)

Can you help to explain it?

bwrsandman:
We have no control over how odoo/master and odoo/8.0 evolve, however, you can clearly see it is behind and not maintained
65 commits behind

>  This branch is 2 commits ahead, 65 commits behind master

Last commit in the runbot module for 8.0: 9 months ago
Last commit in the runbot module for master: 3 days ago

Odoo don't follow OCA rules. We still need to use the most recent runbot. Unless we fork

moylop260:
@bwrsandman 
I got it
Thank you for your explain.

IMHO we have the same case with odoo/odoo 6.1 now we don't have last commit of 8.0 or master in 6.1 but that don't means that I have that use my 6.1 custom modules with instance 8.0 or master.
I have my modules of 6.1 and my modules with 8.0 and my modules with master.
I don't understand this new rule just for odoo-extra.

FYI I tested [odoo-extra/master](https://github.com/odoo/odoo-extra/tree/master) with odoo/odoo 8.0 and I have new errors with other modules.

My proposal is create a new branch called `master` in runbot-addons similar to all other projects and workflow.

moylop260:
The irony is that runbot works with the name of the branch also to build the instances.
But the project where we works runbot-addons we don't like use this standard.
Sorry, but this makes me short circuit in my mind.

@dreispt @gurneyalex What is your opinion in this case?

nhomar:
2015-08-03 19:02 GMT-05:00 moylop260 notifications@github.com:

> @dreispt https://github.com/dreispt @gurneyalex
> https://github.com/gurneyalex What is your opinion in this case?

I am with your @moylop260 we should mantain the standard there.

odoo use the master series with saas-x version and probably with a private
master (or the public one I do not know) it is a good question to Odoo.

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

bwrsandman:
@moylop260 The problem you're getting is due to the other modules. I don't remember where it was discussed, but you can see an explanation in the travis file.

moylop260:
@bwrsandman 
Ok, I have a proposal.
Can you give me your approbal with job_15_pylint to work with both version (8.0 and master)?
For compatibility with other people who choose to use the 8.0 branch (I'm using 8.0 yet)

bwrsandman:
It's not ideal, but you have my :+1:

moylop260:
Thank you man

PR #49, opened 2015-07-27
allow to pass on another repository and database name to custom script

lepistone:
The path to the other repo can be passed on to the custom build script.
For example, this can used to tell a buildout to make a local clone of
odoo instead of cloning everything every time.

### DIFF
--- a/runbot_build_instructions/README.rst
+++ b/runbot_build_instructions/README.rst
@@ -1,0 +1,66 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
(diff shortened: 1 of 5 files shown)


bwrsandman:
Thanks, the change looks :+1: 

Could you elaborate more on how this can be used with a practical example in the module description, please?

lepistone:
I added a better description as suggested by @bwrsandman and another parameter `build_dest` than can be used to pass on the database name to a custom script.

pedrobaeza:
:+1: code review

I think we can move on with this. Merging...

guewen:
Thanks :roller_coaster:

Issue #38, opened 2015-06-04
Challenging contribution: runbot_docker


bwrsandman:
A new module which changes the main test functions of original runbot.

Overload all functions dealing with `runbot/static/*` with the exception of the logs.

An option for a base DockerFile image to simplify the build should be added to the `runbot.repo`.

All builds should use the base DockerFile and clone and merge pull requests. In the job_00.

The job_10 should be based off the job_00 and job_20 should depend on job_10.

The cleanup mechanism should be transposed to clean docker images.

Find a way to optimize `repo.branch` docker files. For example, a PR on 8.0 should start with the 8.0 job_00 image and add a git fetch and merge from the PR source.

Running the docker images should forward the same port inside the instance, but map it the same way it does now (in a range in 2000).

lepistone:
Thanks for writing down the proposal @bwrsandman!

The original runbot is a very specific tool for the needs of Odoo SA. They use it to run the tests, which I don't care about because travis & co already exist. On the other hand, it builds instances that can be tested online, which is what we do for OCA.

With `runbot_build_instructions` and #51, we can have a generic build phase, ideally the same we use on develop, staging and production.

If we add handling of docker images like you suggest here, runbot could in theory become completely independent of odoo: whatever build system like buildout of whatever could produce a docker image, that we deploy on runbot as a multiple staging server of sorts. Then I'd like to be able to deploy the same docker images to normal servers. The different ports can be exposed via docker instead of passing manually options to odoo. At this point we'd have a nice web interface to deploy any multiple docker instances, exposing ports, at every commit/pull request. This is nice, but makes me wonder whether such a thing exists already, and if not, then it would make sense to make the effort independently of odoo.

Does that make any sense? Thanks for your time!

bwrsandman:
Technically, with `runbot_build_instructions` you can test things that aren't odoo.
Are you talking about making runbot independent of Odoo as the server and not the product being tested?
If that's the case, that would require a major rewrite of the logic (mainly the ORM). I think a lot could be gained since Odoo is already quite heavy, has many outdated dependencies and limitations.

It would of course mean a fork and that these addons would no longer benefit from Odoo's improvements to the base runbot. 

Forking is costly, and would only work if there were a lot of contributors and maintainers. I know base runbot isn't the top priority for development at Odoo, but they do have a lot more resources than the OCA has.

moylop260:
FYI You can re-use the .travis.yml to auto-generate the Dockerfile.
You can check this case
[...]

PR #51, opened 2015-08-19
Runbot skip test jobs

lepistone:
To avoid conflicts I included #49 here. The diff will look correct here once #49 is merged. Thanks!

### DIFF
--- a/runbot_build_instructions/runbot_build.py
+++ b/runbot_build_instructions/runbot_build.py
@@ -30,6 +30,7 @@ from openerp.osv import orm, fields
 from openerp.addons.runbot.runbot import mkdirs
 
 _logger = logging.getLogger(__name__)
+MAGIC_PID_RUN_NEXT_JOB = -2
 
 
 def custom_build(func):
@@ -72,7 +73,7 @@ class runbot_build(orm.Model):
     def job_10_test_base(self, cr, uid, build, lock_path, log_path):
         if build.branch_id.repo_id.skip_test_jobs:
             _logger.info('skipping job_10_test_base')
-            return -2
+            return MAGIC_PID_RUN_NEXT_JOB
         else:
             return super(runbot_build, self).job_10_test_base(
                 cr, uid, build, lock_path, log_path
@@ -84,7 +85,7 @@ class runbot_build(orm.Model):
             with open(log_path, 'w') as f:
                 f.write('consider tests as passed: '
                         '.modules.loading: Modules loaded.')
-            return -2
+            return MAGIC_PID_RUN_NEXT_JOB
         else:
             return super(runbot_build, self).job_20_test_all(
                 cr, uid, build, lock_path, log_path
--- a/runbot_build_instructions/runbot_build.py
+++ b/runbot_build_instructions/runbot_build.py
@@ -69,6 +69,27 @@ class runbot_build(orm.Model):
         build.prebuilt = True
         return res
 
+    def job_10_test_base(self, cr, uid, build, lock_path, log_path):
+        if build.branch_id.repo_id.skip_test_jobs:
+            _logger.info('skipping job_10_test_base')
+            return -2
+        else:
+            return super(runbot_build, self).job_10_test_base(
+                cr, uid, build, lock_path, log_path
+            )
+
+    def job_20_test_all(self, cr, uid, build, lock_path, log_path):
+        if build.branch_id.repo_id.skip_test_jobs:
+            _logger.info('skipping job_20_test_all')
+            with open(log_path, 'w') as f:
+                f.write('consider tests as passed: '
+                        '.modules.loading: Modules loaded.')
+            return -2
+        else:
+            return super(runbot_build, self).job_20_test_all(
+                cr, uid, build, lock_path, log_path
+            )
+
     def sub_cmd(self, build, cmd):
         if not cmd:
             return []
--- a/runbot_build_instructions/README.rst
+++ b/runbot_build_instructions/README.rst
@@ -24,6 +24,8 @@ checked:
   build commands. A practical example of that is passing on a repository
   containing a clone of odoo to the custom build script. That script can for
   example make a local clone of odoo, saving both time and disk space.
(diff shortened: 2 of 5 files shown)


pedrobaeza:
Please rebase to let only relevant commits

pedrobaeza:
Anyways seeing the 2 commits, :+1:

max3903:
:+1:

nhomar:
@lepistone We triued this development, but pass as parameters this information, how ensure you are not opening all the server to the user interface?

I do not see any virtualenv (may be I miss something) or ny container....

For example if it allows me pass paramenters I can pass `--stop-after-init | rm -rf ~/` and delete all the home for runbot user making this a huge security flag don't you think?

Did you verify that, may be my code per review was not deep enought... regards.

bwrsandman:
It's a known security issue and it exists even in odoo runbot. Anyone who changes the executable in a pull request can run whatever command they want.
My proposed solution for this https://github.com/OCA/runbot-addons/issues/38

nhomar:
That's true but this open a second security hole didn't it?

Your comment is regarding to "what a PR can make" I understand and I agree, but this is (my question) what a command setted can do.

They are security hole because the same reason but different ones, and we are introducing such error.

guewen:
@nhomar You are right, it is a risk if anyone can access to the configuration of the builds, still this is not a problem in some use cases (if only you have access to the configuration / runbot instance is not shared). 
It seems that #38 would correct the 2 problems. 
The commit your are talking about is already merged in `8.0`, this PR is about the "skip test" (and needs a rebase that's why we see the other commits). So what do you want to do about that? Should we add a disclaimer in the manifest of the module?

bwrsandman:
@nhomar anyone who can add harmful code to the build instructions needs to be either admin or part of the runbot group.

On top of that, adding bash expressions `||` for example will not work since the mechanism here is python system and those work with single commands. Python itself does the piping and the logical operations on return codes, but this is not part of the build_instructions mechanism.

That's not to say the executable couldn't be changed to something harmful, but that is the same issue as with the vanilla runbot. The vanilla runbot security issue is more serious because anyone who can do a pull request can change the code that is run, while build_instructions one would be limited with those who have been granted the rights.

I see them as the same security issue since they're both affected by the more serious issue or the pullrequester being able to change the code that is run. Therefore there needs to be sandboxing such as dockerization.

runbot_build_instructions/runbot_build.py
```diff
@@ -69,6 +69,27 @@ def job_00_init(self, cr, uid, build, lock_path, log_path):
         build.prebuilt = True
         return res
 
+    def job_10_test_base(self, cr, uid, build, lock_path, log_path):
+        if build.branch_id.repo_id.skip_test_jobs:
+            _logger.info('skipping job_10_test_base')
+            return -2
```
bwrsandman: Can you explain this magic number?

lepistone: @bwrsandman https://github.com/odoo/odoo-extra/blob/master/runbot/runbot.py#L1015-L1018

lepistone:
As discussed, I don't think we're making the vulnerability worse. While anyone can make a public pull request, only an administrator can change the runbot settings. I think this PR can go forward independently of #38.

Thanks!

runbot_build_instructions/runbot_build.py
pedrobaeza: Please put a note explaining that this is the expected number in runbot for directly calling next job

bwrsandman: Even better: factor it as a global variable, that way it's self documenting and if it changes in the future, it can be fixed at one place.

bwrsandman:
@lepistone Agreed

nhomar:
A disclaimer in the module is enough at least until we find and or develop
a proper and more secure way.

On Thursday, August 20, 2015, Guewen Baconnier notifications@github.com
wrote:

> @nhomar https://github.com/nhomar You are right, it is a risk if anyone
> can access to the configuration of the builds, still this is not a problem
> in some use cases (if only you have access to the configuration / runbot
> instance is not shared).
> It seems that #38 https://github.com/OCA/runbot-addons/issues/38 would
> correct the 2 problems.
> The commit your are talking about is already merged in 8.0, this PR is
> about the "skip test" (and needs a rebase that's why we see the other
> commits). So what do you want to do about that? Should we add a disclaimer
> in the manifest of the module?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/runbot-addons/pull/51#issuecomment-132920955.

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

bwrsandman:
That could be added to the module's description, but I would respectfully ask you to do this in another pull request. This one is on the affected module, but the changeset not related and we shouldn't be tagging on feature requests until we end up patchbombing.

nhomar:
2015-08-20 10:46 GMT-05:00 Sandy notifications@github.com:

> That could be added to the module's description, but I would respectfully
> ask you to do this in another pull request. This one is on the affected
> module, but the changeset not related and we shouldn't be tagging on
> feature requests until we end up patchbombing.

Agreed also.

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

moylop260:
:+1:

runbot_build_instructions/runbot_build.py
lepistone: done

guewen:
The fail is due to https://github.com/odoo/odoo-extra/issues/60
See also https://github.com/OCA/runbot-addons/pull/54#issuecomment-133386542

guewen:
:+1:

PR #40, opened 2015-06-04
Fix runbot gitlab

angelmoya:
As described in issue [https://github.com/OCA/runbot-addons/issues/35]
I check if repository is set as gitlab repository before call _set_gitlab_ci_conf_
Also fix pep8, reestructure files, and set description in readme.rst

PR #52, opened 2015-08-20
[IMP] Run set_gitlab_ci_conf only when uses_gitlab is true

PCatinean:


pedrobaeza:
https://github.com/OCA/runbot-addons/pull/40 already does this, and contains more improvements, please check it instead of this one.

PR #53, opened 2015-08-20
fix deletion of our modules (courtesy @moylop260)

lepistone:


pedrobaeza:
:+1:

moylop260:
This fixed next iissue of empty folder of runbot-addons:
https://youtu.be/r6-IE87UZQc

:+1:

yvaucher:
:+1:

lepistone:
Ok, that didn't delete the module at all. Hurm.

lepistone:
Amended. Let's see if that's better. (I was deleting `website_twitter` instead of `odoo-extra/website-twitter`).

.travis.yml
```diff
@@ -33,7 +33,7 @@ install:
   - travis_install_nightly
   - git clone --depth=1 https://github.com/odoo/odoo-extra ~/odoo-extra
   # odoo-extra has a bunch of v9 modules which aren't compatible, remove them
-  - rm -rf $(ls | grep -v runbot$)
+  - (cd ~/odoo-extra; rm -rf $(ls | grep -v runbot$))
```
bwrsandman, line 36: Are those outer parenthesis necessary?

lepistone:
This has revealed that `runbot_janitor` tests fail (none were run). Since the failure is now real, I think we can merge this one.

pedrobaeza:
OK, merging. About parenthesis, I think it's not very important if they're redundant because the expression is still very readable

bwrsandman:
@lepistone janitor is not used by anyone, it would be safe to remove it

.travis.yml
lepistone, line 36: they make a subshell, so there is no need to `cd ..` afterwards.

lepistone:
@bwrsandman #54 then. Thanks

PR #54, opened 2015-08-20
remove runbot_janitor because tests fail

lepistone:
If someone wants to adopt it and fix the tests, feel free of course.
Thanks!


pedrobaeza:
It's still failing...

nhomar:
will we drop modules instead fix them? there is a deeper explanation that the test fail? I feel we should finish it, and only mark it as installable false to avoid lose history.

bwrsandman:
I wrote this module for my own purposes. It never did what it was supposed to do and its function has changed. It never worked and has been abandoned. It also provides no real function that runbot doesn't do on its own.

@lepistone is removing it at my request.

bwrsandman:
:+1:

nhomar:
@lepistone @bwrsandman Ok excellent, :+1: in that case.!

moylop260:
:+1:

moylop260:
FYI the original runbot have the error of 

``` python
AccessError: ('AccessError', u'Sorry, you are not allowed to access this document. Only users with the following access level are currently allowed to do that:\n- Runbot/User\n\t- Runbot/Manager\n\n(Document model: runbot.repo)')
```

Then this PR can be merged

lepistone:
Thanks @moylop260. I see we still we have this `AccessError` on a "crawler test" (not a test from `runbot-addons`). Do you know anything else about it?

moylop260:
Yeah, runbot module original added groups and security but this create a error in test.
Odoo sa dont have a runbot to check runbot self.

moylop260:
https://github.com/odoo/odoo-extra/issues/60
Issue reported 3 months ago

lepistone:
thanks for the info @moylop260

PR #57, opened 2015-08-26
[FIX] Fixed method call to get closest branch name

PCatinean:


runbot_pylint/models/runbot_build.py
```diff
@@ -91,8 +91,8 @@ def get_repo_branch_name(self):
         for build in self:
             hint_branches = set()
             for extra_repo in build.repo_id.dependency_ids:
-                closest_name = build.get_closest_branch_name(
-                    extra_repo.id, hint_branches)
+                repo_id, closest_name = build._get_closest_branch_name(
+                    extra_repo.id)
```
gurneyalex: which branch of odoo-extra are you using as reference for this method rename?

PCatinean: The latest version of master, the function call is the same as: 
https://github.com/odoo/odoo-extra/blob/master/runbot/runbot.py#L732

pedrobaeza:
Waiting for the answer to review easily...

gurneyalex:
:+1:

runbot_pylint/models/runbot_build.py
pedrobaeza: If you're not using repo_id, make `closest_name = build._get_closest_branch_name(extra_repo.id)[1]`

pedrobaeza:
:+1: Thanks, merging...

Issue #61, opened 2015-09-01
ResourceNotFound Error with runbot_gitlab


npiganeau:
Hi,

I use gitlab 7.14.1 (docker image) with runbot (Odoo v8 + runbot master from odoo-extra).

The runbot works perfectly for building branches from gitlab without gitlab integration with runbot_gitlab module.

However, when I install runbot_gitlab and set "use gitlab" option I get the following error:

```
runbot_1 | 2015-08-31 17:21:23,393 53 ERROR runbot openerp.http: Exception during JSON request handling.
runbot_1 | Traceback (most recent call last):
runbot_1 |   File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 530, in _handle_exception
runbot_1 |     return super(JsonRequest, self)._handle_exception(exception)
runbot_1 |   File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 567, in dispatch
runbot_1 |     result = self._call_function(**self.params)
runbot_1 |   File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 303, in _call_function
runbot_1 |     return checked_call(self.db, *args, **kwargs)
runbot_1 |   File "/usr/lib/python2.7/dist-packages/openerp/service/model.py", line 113, in wrapper
runbot_1 |     return f(dbname, *args, **kwargs)
runbot_1 |   File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 300, in checked_call
runbot_1 |     return self.endpoint(*a, **kw)
runbot_1 |   File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 796, in __call__
runbot_1 |     return self.method(*args, **kw)
runbot_1 |   File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 396, in response_wrap
runbot_1 |     response = f(*args, **kw)
runbot_1 |   File "/usr/lib/python2.7/dist-packages/openerp/addons/web/controllers/main.py", line 953, in call_button
runbot_1 |     action = self._call_kw(model, method, args, {})
runbot_1 |   File "/usr/lib/python2.7/dist-packages/openerp/addons/web/controllers/main.py", line 941, in _call_kw
runbot_1 |     return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
runbot_1 |   File "/usr/lib/python2.7/dist-packages/openerp/api.py", line 241, in wrapper
[...]

PR #62, opened 2015-09-02
[FIX] Calling branch method of gitlab API without name kwarg

PCatinean:
Calling branch method of the Gitlab Project with keyword argument name fails to retrieve branch info

Issue was also referenced by other users and the fix seemed to work https://github.com/OCA/runbot-addons/issues/61#issuecomment-136794670

Method definition in API class:

https://github.com/alexvh/python-gitlab3/blob/master/gitlab3/_api_definition.py#L257

### DIFF
--- a/runbot_gitlab/runbot_repo.py
+++ b/runbot_gitlab/runbot_repo.py
@@ -201,7 +201,7 @@ class RunbotRepo(models.Model):
             source_project = get_gitlab_project(
                 self.base, self.token, mr.source_project_id
             )
-            source_branch = source_project.branch(name=mr.source_branch)
+            source_branch = source_project.branch(mr.source_branch)
             commit = source_branch.commit
             sha = commit['id']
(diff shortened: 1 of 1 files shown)


gurneyalex:
:+1: (no test)

guewen:
:+1:

pedrobaeza:
Why Travis is red?

PCatinean:
Unfortunately it has been like this for all PR's, nothing to do with the changes themselves, it seems it's some sort of Access Error: AccessError: ('AccessError', u'Sorry, you are not allowed to access this document. Only users with the following access level are currently allowed to do that:\n- Runbot/User\n\t- Runbot/Manager\n\n(Document model: runbot.repo)')

guewen:
Upstream bug: https://github.com/odoo/odoo-extra/issues/60

pedrobaeza:
:+1:

PR #60, opened 2015-08-31
Add multihosting modules

niboo-sam:
Add multihosting modules that allows to use different repository hosting

PR #64, opened 2015-09-02
[IMP] Add translations folder and two extra boolean fields to runbot_gitlab

PCatinean:
Standard behavior of module is:
1. To set all protected branches as sticky 
2. To build only merge requests and sticky branches while ignoring normal branches

Users (such as myself) might want a different behavior on this and there should be flexibility allowing this change

As such I made two boolean fields managing the desired behavior and

### DIFF
--- a/runbot_gitlab/runbot_repo.py
+++ b/runbot_gitlab/runbot_repo.py
@@ -76,6 +76,7 @@ def gitlab_api(func):
     """Decorator for functions which should be overwritten only if
     uses_gitlab is enabled in repo.
     """
+
     def gitlab_func(self, *args, **kwargs):
         if self.uses_gitlab:
             return func(self, *args, **kwargs)
@@ -146,7 +147,18 @@ def set_gitlab_ci_conf(token, gitlab_url, runbot_domain, repo_id):
 
 class RunbotRepo(models.Model):
     _inherit = "runbot.repo"
-    uses_gitlab = fields.Boolean('Use Gitlab')
+
+    uses_gitlab = fields.Boolean(string='Use Gitlab')
+
+    mr_only = fields.Boolean(
+        string="MR Only",
+        default=True,
+        help="Build only merge requests and skip regular branches")
+
+    sticky_protected = fields.Boolean(
+        string="Sticky for Protected Branches",
+        default=True,
+        help="Set all protected branches on the repository as sticky")
 
     @api.model
     def create(self, vals):
@@ -188,6 +200,9 @@ class RunbotRepo(models.Model):
     @api.one
     @gitlab_api
     def update(self):
+
+        branch_obj = self.env['runbot.branch']
+
         project = get_gitlab_project(self.base, self.token)
 
         merge_requests = project.find_merge_request(
@@ -221,7 +236,7 @@ class RunbotRepo(models.Model):
             subject = commit['message']
             title = mr.title
             # Create or get branch
-            branch_ids = self.env['runbot.branch'].search([
+            branch_ids = branch_obj.search([
                 ('repo_id', '=', self.id),
                 ('project_id', '=', project.id),
                 ('merge_request_id', '=', mr.iid),
@@ -231,7 +246,7 @@ class RunbotRepo(models.Model):
             else:
                 logger.debug('repo %s found new Merge Proposal %s',
                              self.name, title)
-                branch_id = self.env['runbot.branch'].create({
+                branch_id = branch_obj.create({
                     'repo_id': self.id,
                     'name': title,
                     'project_id': project.id,
@@ -266,7 +281,7 @@ class RunbotRepo(models.Model):
             state='closed'
         ))
 
-        closed_mrs = self.env['runbot.branch'].search([
+        closed_mrs = branch_obj.search([
             ('merge_request_id', 'in', closed_mrs),
         ])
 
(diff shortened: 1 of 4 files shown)


runbot_gitlab/runbot_repo_view.xml
```diff
@@ -9,6 +9,8 @@
 
         <field name="token" position="before">
```
pedrobaeza: You can put all Gitlab related options under a new group

runbot_gitlab/runbot_repo.py
```diff
@@ -146,7 +147,16 @@ def set_gitlab_ci_conf(token, gitlab_url, runbot_domain, repo_id):
...
-    uses_gitlab = fields.Boolean('Use Gitlab')
+
+    uses_gitlab = fields.Boolean(string='Use Gitlab')
+
+    mr_only = fields.Boolean(
+        string="MR Only",
+        help="Build only merge requests and skip regular branches")
+
+    sticky_protected = fields.Boolean(
+        string="Sticky Protected Branches",
```
pedrobaeza: Better "Sticky for Protected Branches"

runbot_gitlab/runbot_repo.py
```diff
@@ -279,25 +292,29 @@ def update(self):
...
-        protected_branches.add(project.default_branch)
+        if self.sticky_protected:
+            # Put all protected branches as sticky
+            protected_branches = set(
+                b.name for b in project.find_branch(find_all=True, protected=True)
+            )
+            protected_branches.add(project.default_branch)
 
-        for branch in self.env['runbot.branch'].search([
+            sitcky_protected_branches = branch_obj.search([
```
pedrobaeza: s/sitcky/sticky

runbot_gitlab/runbot_repo.py
```diff
@@ -279,25 +292,29 @@ def update(self):
...
-            ('repo_id', 'in', [i.id for i in self]),
-            ('project_id', '=', False),
-            ('merge_request_id', '=', False),
-        ])
-        for build in self.env['runbot.build'].search([
-                ('branch_id', 'in', [b.id for b in branches])]):
-            build.skip()
+            ])
+
+            for branch in sitcky_protected_branches:
```
pedrobaeza: `sticky_protected_branches.write({'sticky': True})`

runbot_gitlab/runbot_repo.py
```diff
@@ -279,25 +292,29 @@ def update(self):
...
+                branch.write({'sticky': True})
+
+        if self.mr_only:
+            # Skip non-sticky non-merge proposal builds
+            branches = branch_obj.search([
+                ('sticky', '=', False),
+                ('repo_id', 'in', [i.id for i in self]),
```
pedrobaeza: `('repo_id', 'in', self.ids),`

runbot_gitlab/runbot_repo.py
```diff
@@ -279,25 +292,29 @@ def update(self):
...
+                ('project_id', '=', False),
+                ('merge_request_id', '=', False),
+            ])
+            for build in self.env['runbot.build'].search([
+                    ('branch_id', 'in', [b.id for b in branches])]):
```
pedrobaeza: `branches.ids`

runbot_gitlab/runbot_repo.py
```diff
@@ -146,7 +147,16 @@ def set_gitlab_ci_conf(token, gitlab_url, runbot_domain, repo_id):
 
 class RunbotRepo(models.Model):
     _inherit = "runbot.repo"
-    uses_gitlab = fields.Boolean('Use Gitlab')
+
+    uses_gitlab = fields.Boolean(string='Use Gitlab')
```
hbrunn, line 151: I lately discovered https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field (third bullet point) - this applies to this and the next field

PCatinean, line 151: That's very interesting hbrunn

The initial thought was that "explicit is always better than implicit"
So in conclusion for fields that have the exact same name as the variable omit the string parameter (both positional and keyword argument) ?

PCatinean:
Using a separate group produces this result

runbot-gitlab

The rest of the fixes are in place

pedrobaeza:
I was suggesting to put string="Gitlab" to the group to have as a separate section. Anyway, perhaps a better option is the approach of https://github.com/OCA/runbot-addons/pull/60 and add gitlab support via this "framework"

runbot_gitlab/runbot_repo.py
PCatinean, line 151: I have tested this just now and apparently it just capitalizes the first word and the other not, perhaps that file should be updated

pedrobaeza:
Squash your commits in 1 and you have my :+1:

pedrobaeza:
:+1:

runbot_gitlab/runbot_repo.py
guewen, line 151: Yes, it capitalizes only the first word, so personally I omit `string` only when the name is a single word.

guewen:
:+1:

runbot_gitlab/runbot_repo.py
```diff
@@ -146,7 +147,16 @@ def set_gitlab_ci_conf(token, gitlab_url, runbot_domain, repo_id):
...
+
+    mr_only = fields.Boolean(
+        string="MR Only",
+        help="Build only merge requests and skip regular branches")
```
StefanRijnhart, line 156: By not setting the default to True, you are changing the default behaviour of this module. Existing uses might be confused. Would you consider setting the defaults so as to preserve the previous behaviour?

runbot_gitlab/runbot_repo.py
```diff
@@ -146,7 +147,16 @@ def set_gitlab_ci_conf(token, gitlab_url, runbot_domain, repo_id):
...
+
+    sticky_protected = fields.Boolean(
+        string="Sticky for Protected Branches",
+        help="Set all protected branches on the repository as sticky")
```
StefanRijnhart, line 161: Consider setting default to True here as well.

runbot_gitlab/runbot_repo.py
PCatinean, line 156: That is actually a very good point indeed, updating now
Thanks!

gurneyalex:
:+1:

Can you update the translation files?

runbot_gitlab/runbot_repo.py
```diff
@@ -279,25 +294,28 @@ def update(self):
...
 
-        # Put all protected branches as sticky
-        protected_branches = set(
-            b.name for b in project.find_branch(find_all=True, protected=True)
-        )
-        protected_branches.add(project.default_branch)
+        if self.sticky_protected:
+            # Put all protected branches as sticky
+            protected_branches = set(
+                b.name for b in project.find_branch(find_all=True, protected=True)
```
StefanRijnhart: While tests are broken for other reasons as well, flake8 does signal that this line is too long.

PR #73, opened 2015-12-09
[UPD] update to gitlab >= 8.1, major refactoring

hbrunn:
I dropped support for gitlab <= 8.0. As the new CI API is not supported by gitlab3, I dropped this dependency too.

This works with https://github.com/odoo/odoo-extra/tree/master and https://github.com/odoo/odoo-extra/tree/8.0


hbrunn:
read https://github.com/hbrunn/runbot-addons/blob/8.0-runbot_gitlab-8.1-ci-api/runbot_gitlab/README.rst for an introduction on how to use this

npiganeau:
@hbrunn please have a look at my implementation at https://github.com/ndp-systemes/runbot-addons with the runbot_gitlab_new module.

I was about to submit a PR with it when I saw yours. I think we should merge our efforts on this.

My solution is similar to yours, but:
- no need to setup builds, .gitlab-ci.yml, etc. as it uses gitlab api (not ci) so builds remain available for other testing if needed + works same way as github
- MR are fetched by git directly through the new refs/merge-requests/*

Works in production here.

hbrunn:
what a pity you didn't propose your code earlier. The code proposed here either uses builds or MRs, so if you don't want to use builds, you can leave them out and just keep working with MRs. The beauty of the builds is that you get feedback directly within gitlab.

There's still a problem with MRs and builds together, but that should be solvable/

npiganeau:
Of course I get the status of runbot builds within gitlab (that's the point actually) without gitlab builds. The only thing I don't have is the trace in gitlab but I have the link to runbot so same number of clicks + I can login into the build from runbot which I couldn't from gitlab.

Actually it works exactly the same way as it does with github.

hbrunn:
@npiganeau thanks for the hint, the commit status updates slipped me in the API documentation (or it wasn't there back then, no idea). I eventually got frustrated with the build API as it changed during the last minor releases radically. Now I do something very similar to your code

runbot_gitlab/models/runbot_build.py
```diff
@@ -0,0 +1,90 @@
...
+from .runbot_repo import escape_branch_name
+
+
+class runbot_build(models.Model):
+    _inherit = "runbot.build"
+
+    @api.depends('repo_id.uses_gitlab', 'branch_id', 'name')
+    def _compute_dest(self):
+        for build in self.filtered('repo_id.uses_gitlab'):
+            nickname = escape_branch_name(build.branch_id.name)
```
tremlin: I would suggest to truncate the `nickname` after 32 chars as in the original code because too long names can cause trouble with the database and the filesystem.

runbot_gitlab/models/runbot_repo.py
```diff
@@ -0,0 +1,335 @@
...
+                        'merge_request_id':
+                        branch.get('merge_request', {}).get('iid'),
+                    })
+                # in this case, this branch's builds were skipped, undo that
+                if self.mr_only:
+                    self.env['runbot.build'].search([
+                        ('branch_id', 'in', branch_ids.ids),
+                        ('result', '=', 'skipped'),
+                        ('state', '=', 'done'),
+                    ], limit=1).force()
```
tremlin, line 244: Probably I don't understand the whole updating logic, but isn't there a chance that this line would also force the build of no-MR branches?

tremlin:
@hbrunn @npiganeau Thanks for your efforts!

I reviewed the code and I think it's almost ready for merging. I've also installed the new module in a test environment and will give you some feedback from that soon.

hbrunn:
thanks @tremlin for your comments, if your testing results in any code changes, I'll be more than happy to integrate them in this PR, this way, we should be able to finish this hopfully

tremlin:
I've been testing this module for one week with a test Odoo and a production Gitlab: it works very well. The only thing that I would really like to have is the truncation of the nickname to 32 characters (see code comment above) like in the original `runbot` module. If the nickname is too long, it will cause problems at least with PostgreSQL where the database name is limited by 64 characters. Is the missing truncation on purpose?

hbrunn:
not by me, just PR into https://github.com/hbrunn/runbot-addons/tree/8.0-runbot_gitlab-8.1-ci-api what you propose how it should be

hbrunn:
BTW: the 8.0 branch of https://github.com/odoo/odoo-extra is heavily outdated, you should use the master branch. Possibly there runbot itself already does sanitizing so that it's not necessary here (on our runbot this works completely fine). Which runbot version do you use?

tremlin:
I use the master branch of runbot. I've created a PR to your repository together with a link to the original line that I am referring to.

If you have only reasonably short branch names (less than 53 characters), you also won't encounter problems without truncation :)

hbrunn:
okay, clear. Thanks for you PR, merged

tremlin:
Great! I reviewed and tested the code, :+1: for merging.

hbrunn:
cool. now we just need a second thumbs up...

pedrobaeza:
Making a non exhaustive code review, :+1: 

More testing will come later on production.

OSguard:
I was currently testing on @tremlin / our instance .. investigating small issues, so as this is already merge you get a issue or ar PR later ;)
thanks for your work

PR #74, opened 2015-12-12
[ADD] runbot_website_display

hbrunn:

PR #76, opened 2016-01-15
[ADD] ADD module runbot_repository_visibility

jjscarafia:
Add module runbot_repository_visibility that allows you to order repositories and choose which of them you want to be visible on website

### DIFF
--- a/runbot_repository_visibility/README.rst
+++ b/runbot_repository_visibility/README.rst
@@ -1,0 +1,42 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Runbot Repository Visibility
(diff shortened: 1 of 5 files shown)


hbrunn:
@jjscarafia it looks like the two of us had exactly the same idea: https://github.com/OCA/runbot-addons/pull/74

nhomar:
@hbrunn Agreed they are both almost the same the unique difference is that @jjscarafia inherited more templates (may be because of overwork and or considering more views which you dont), @jjscarafia can you join your work with @hbrunn which is older?

jjscarafia:
Thanks you both!
I close this PR because it does exactly the same as this #74. The only difference is that I inherit two q-web views to limit the sort and #74 inherits the controller.

PR #81, opened 2016-02-29
[IMP] runbot_gitlab: fix minor issues

tremlin:
This pull request addresses two minor issues:
1. It seems that Gitlab expects just the branchname without leading 'refs/...' for the commit status API. With the complete name the link to the build in our Gitlab instance is broken.
2. The nickname should also be converted to lowercase. This is because dbfilter is case-sensitive. If we  don't convert uppercase characters, there is a mismatch between the hostname, which is treated case-insensitively by the browser, and the database name.


hbrunn:
:+1:

pedrobaeza:
:+1:

OSguard:
:+1: testet

PR #84, opened 2016-03-09
[FIX] use relative urls

hbrunn:
This is quite similar to https://github.com/odoo/odoo-extra/pull/90 - we shouldn't make assumptions about where runbot runs and how it is accessible


pedrobaeza:
:+1:

max3903:
:+1:

gurneyalex:
:+1:

PR #87, opened 2016-07-08
[MIG] runbot_secure: Upgrade to v9

lasley:
This brings runbot_secure to 9.0. Pretty simple conversion, just OCA style stuff.


moylop260:
👍 
CI red unrelated

elicoidal:
:+1:

hbrunn:
Our runbot runs completely on SSL, which works because of a bunch of patches that make all links (protocol-)relative: https://github.com/odoo/odoo-extra/pull/90/files
Wouldn't it make sense to try harder to get this merged, then we wouldn't need this module at all?

lasley:
I agree. This module actually doesn't hit all the links either. 

That said- I have outright given up on patches to anything in an Odoo SA repo. Do you think there is a chance they will actually merge your PR?

lasley:
@hbrunn After looking at your code for a bit, I think I may be able to just modify this module to apply the relative URIs instead. Assuming that is a valid solution, I can modify this PR and update the naming of the module to maybe runbot_relative_uri or something.

If you think that we can get Odoo SA to merge something community, then I am with you on that being a better solution. I'm honestly at the point of forking my own Odoo repos though, because there's some outright bugs that they just refuse to fix or accept fixes for. That's a conversation for a different day though 😉

pedrobaeza:
@lasley, we have OCB for that: https://github.com/OCA/OCB :wink:

lasley:
@pedrobaeza Interesting- I've always wondered what that repo was, but I figured it was some sort of LTS type repo for supporting things that Odoo SA did away with (purely by the naming backport). We are also using this for new features & bugs that Odoo SA isn't fixing?

lasley:
An idea in OCA/connector-telephony#90 comes to mind in terms of things I want in core - or at least to be able to entertain the idea of it being there. 

This would be awesome if it would actually be possible with OCB. I think we maybe need a way to let more of the community know about this.

pedrobaeza:
Well, maybe it's not exactly what you are thinking of. This is only for bugfixes that Odoo refuses to include. Changes that made OCB to not be compatible with Vanilla Odoo are not allowed for now. Anyway, with the inheritance possibilities, almost everything can be plug via a module (although it can be simpler having it in core, I know).

lasley:
Understood, thanks for the elaboration. I have a bug fix I was about to submit to Odoo SA that I will submit there too.

You're definitely correct about the modularity, one of the reasons why I love Odoo so deeply. I'll hold back further discussions on a forward-development core fork until I have a solid use case where I can prove something to be significantly less efficient, or downright impossible in a module 😉

hbrunn:
coming back to the original topic: I think your proposal with sticking my PR into a module to replace this one is the best way to go. At least my PRs for runbot (https://github.com/odoo/odoo-extra/pulls/hbrunn) were never merged.

lasley:
I'm for it. I'll see if I also get the links that are missed in this module - the log txt files.

If someone will please tag this WIP, I should have done sometime in the next week or two.

lasley:
That was easy - this is gtg. 

It would be nice if someone has an idea on how to actually manipulate an attribute instead of replacing, or somehow using it in the replacement as seed text. This does work though, albeit a bit sloppy and hard-coded.

hbrunn:
:+1:
In this case, it might be possible to use `<attribute remove="http:" separator="//" />`: https://github.com/OCA/OCB/blob/9.0/openerp/addons/base/ir/ir_ui_view.py#L552

lasley:
Awesome suggestion @hbrunn - this is going to be useful going forward!

For this though, I can't figure out a way to work within what we need and this split/join filter. Below is the test case for your suggestion, but I tried basically every combination I could think of:

```
>>> to_remove = ['http:']
>>> separator = '//'
>>> values = "http://{{bu['domain']}}/?db={{bu['real_dest']}}".split(separator)
>>> value = (separator or ' ').join(filter(lambda s: s not in to_remove, values) + to_add)
>>> value
"{{bu['domain']}}/?db={{bu['real_dest']}}"
```

The issue is that we need an exact match for the remove. Once it is removed, we no longer have that index in the array and then join doesn't get us our beloved relative URI 😦

hbrunn:
hmm, right. Stuff could be so simple if the body of the loop was its own function

lasley:
Snap wow yeah the possibilities if that were the case! It's interesting when looking at Odoo code - you can almost always discern the old from the new based on how extensible it is in places like this.

lasley:
Alright I switched this up to a JS-style override instead of the XML. 

With this strategy, we are now catching all of the appropriate URIs & are compatible with any view schema (Odoo SA and the Vauxoo fork tested)

I would like to create a tour, but that would involve a lot of additional demo and stuff which realistically should be in Runbot core. IMO it's worth just moving forward on this with manual testing, seeing as we're crutched by that basically everywhere else in this repo too.

We're including this in our Runbot image, so it's easily testable from there is anyone has had the balls to switch over. It's already installed on our Runbot and working great.

moylop260:
@lasley Could you update your branch, please?

lasley:
Updated

moylop260:
Could you check the lint build result from travis, please?

lasley:
Lint has been fixed, as well as a bug that I noticed on my Runbot where not all the link were being transformed when there are multiple build agents.

lasley:
Confirmed this now works on all the build/log links in our Runbot. The last implementation missed a few on the index page.

moylop260:
thanks

PR #89, opened 2016-07-08
[REF] runbot_gitlab: extract POST data generation into separate method

tremlin:
This MR refactors the code that sends data to Gitlab for a new build. It moves the generation of the POST data of the request into a separate method so that if can be adapted in custom modules.


hbrunn:
:+1:

PR #91, opened 2016-08-17
[FIX] runbot_build_instructions: use no-xmlrpcs only if it is supported by the server

tremlin:
In the module `runbot_build_instructions` the parameter `no-xmlrpcs` is always used for starting Odoo children. This command-line switch is no longer available in Odoo v9 and the module `runbot_build_instructions` cannot start children.
This pull request uses the same solution as the main `runbot` code and greps the file that defines the command-line options for the string.


hbrunn:
closing due to lack of activity

PR #92, opened 2016-09-27
[REF] runbot_travis2docker: Long term test

moylop260:
[runbot skip to create builds for old branches](https://github.com/odoo/odoo-extra/blob/038fd3ecf02fa6f63c308cfc723ff26e548d7041/runbot/runbot.py#L326-L328) then we need to force create it if we don't have changes in demo repo

### DIFF
--- a/runbot_travis2docker/tests/test_runbot_build.py
+++ b/runbot_travis2docker/tests/test_runbot_build.py
@@ -65,14 +65,12 @@ class TestRunbotJobs(TransactionCase):
             })
         self.assertEqual(len(branch), 1, "Branch not found")
         self.build_obj.search([('branch_id', '=', branch.id)]).unlink()
-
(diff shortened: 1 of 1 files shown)


moylop260:
@pedrobaeza 
What do you think about?

moylop260:
@lasley @alan196
Your feedback is welcome

pedrobaeza:
No opinion from my side, as I don't know this code, sorry.

lasley (approved):
Seems to work on my end

moylop260:
Could we merge it?

lasley:
I'm ok with it. Realistically it would likely be one of us two fixing any issue that arises & the change looks pretty not dangerous IMO

hbrunn:
still relevant?

PR #101, opened 2016-11-02
8.0 runbot website display

hbrunn:
small fixes

### DIFF
--- a/runbot_website_display/models/runbot_repo.py
+++ b/runbot_website_display/models/runbot_repo.py
@@ -9,5 +9,5 @@ class RunbotRepo(models.Model):
 
     _order = 'sequence asc'
 
-    sequence = fields.Integer()
+    sequence = fields.Integer(default=0)
     website_published = fields.Boolean(default=True)
--- a/runbot_website_display/__openerp__.py
+++ b/runbot_website_display/__openerp__.py
@@ -12,6 +12,7 @@
         'runbot',
     ],
     "data": [
+        "views/runbot_branch.xml",
         "views/runbot_repo.xml",
     ],
     "installable": True,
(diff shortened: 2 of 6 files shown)


runbot_website_display/models/runbot_repo.py
```diff
@@ -9,5 +9,5 @@ class RunbotRepo(models.Model):
 
     _order = 'sequence asc'
 
-    sequence = fields.Integer()
+    sequence = fields.Integer(default=0)
```
pedrobaeza, line 12: 0 is already the default, isn't it?

hbrunn: not for existing records it seems

hbrunn: aha, here we go: https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L2394, https://github.com/OCA/OCB/blob/8.0/openerp/fields.py#L425, and there's no "default default" set in the `Integer` class: https://github.com/OCA/OCB/blob/8.0/openerp/fields.py#L1037

Interesting, so we should always set defaults for integer fields it seems, otherwise we get NULLs in the database on installation, which messes up sorting in this case

pedrobaeza: Uhm, that seems something to be patched upstream, don't you think?

pedrobaeza (approved):

runbot_website_display/models/runbot_repo.py
hbrunn: sure, but I'm in the middle of something else right now. If you've some time to spare, go ahead

hbrunn:
sequence already is in upstream, and we don't need the publishing option any more, so I close this

Issue #105, opened 2016-11-07
runbot_travis2docker: does not seem to work on OCA runbot


gurneyalex:
I probably got some configuration wrong. 

I'm using docker 1.6.2 (from jessie-backports), the user running odoo is a member of the docker group (but odoo is not running inside docker). 

Test repository is https://runbot.odoo-community.org/runbot/repo/github-com-vauxoo-dev-runbot-branch-remote-name-grp-feature2-232

Setup:

![screenshot from 2016-11-07 17-39-14](https://cloud.githubusercontent.com/assets/2032807/20066293/58210b50-a511-11e6-889a-22891b8fccd0.png)
I get empty builds or skipped builds, `docker ps` says nothing, and `docker images` is also empty. 

Weird line in the logs:
```
2016-11-07 16:23:12,856 4580 INFO openerp-oca-runbot openerp.addons.runbot_travis2docker.models.runbot_build: Dockerfile without TESTS=1 env. Skipping builds [3271674]
```

gurneyalex:
@moylop260 your feedback is very welcome on this. Thanks for your help.

moylop260:
What is your `travisfile2dockerfile --version`?

moylop260:
`travisfile2dockerfile --version`

lasley:
The Travis file needs to be upgraded to the new version. This error is saying there is no "TESTS=1" in the matrix, which is defaulted so doesn't hurt Travis. 

Happens all the time on our instance, which is why you see me submitting new Travis files quite often. You can see the transition on our [web](https://runbot.laslabs.io/runbot/repo/git-github-com-laslabs-web-git-54) repo. Skipped are all before merge of OCA/web#447 or OCA/web#448

Although @moylop260 comment maybe alludes to this being fixed in a newer T2D version?

moylop260:
@gurneyalex 
I'm using the following `runbot.repo` configuration:
![docker](https://cloud.githubusercontent.com/assets/6644187/20067819/351772aa-a4dc-11e6-8317-36ca42c94a87.png)

It's building here:
http://runbot.vauxoo.com/runbot/repo/https-github-com-vauxoo-dev-runbot-branch-remote-name-grp-feature2-git-90

I don't see the url in the image of your first comment but could change it by:
https://github.com/vauxoo-dev/runbot_branch_remote_name_grp_feature2.git

NOTE: The docker version is too old this could get new errors by old bug but Its is other matter.

moylop260:
[...]

PR #107, opened 2016-11-15
[FIX] runbot_travis2docker: Fix server options

lasley:
* Fix interpolation on server options due to multi line

This should fix log issue reported in #105. Multiline string & `.format` weren't playing nicely, resulting in the following in SERVER_OPTIONS:
```
"--log-db=postgres://{cfg[db_user]}:{cfg[db_password]}@localhost/runbot"
```

### DIFF
--- a/runbot_travis2docker/models/runbot_build.py
+++ b/runbot_travis2docker/models/runbot_build.py
@@ -121,8 +121,10 @@ class RunbotBuild(models.Model):
         ] + pr_cmd_env
         logdb = cr.dbname
         if config['db_host'] and not travis_branch.startswith('7.0'):
-            logdb = 'postgres://{cfg[db_user]}:{cfg[db_password]}@' +\
-                    '{cfg[db_host]}/{db}'.format(cfg=config, db=cr.dbname)
+            logdb = 'postgres://%s:%s@%s/%s' % (
+                config['db_user'], config['db_password'],
+                config['db_host'], cr.dbname,
+            )
         cmd += ['-e', 'SERVER_OPTIONS="--log-db=%s"' % logdb]
         return self.spawn(cmd, lock_path, log_path)
 


moylop260:
Sorry, I don't understand what is wrong with the original [runbot-formating](https://github.com/odoo/odoo-extra/blob/5b29b651ba3fe3604d1ec93870fc98854a023df4/runbot/runbot.py#L896)

Could you help to understand?

lasley:
@moylop260 We added #94 to allow for log-db to work using Travis2Docker. In #94, I used basically the same code you linked from Odoo.

The difference between our's and their's is PEP8. The string is multiline, but the format is only on the last. This PR makes the string itself single line in order for the interpolation to affect all of it instead of just the second part.

lasley:
Contrived example:

```
>>> '{a[test]}'+ \
... '{a[test]}'.format(a={'test':'test val'})
'{a[test]}test val'
```

moylop260:
I got it
Thanks for clarify

runbot_travis2docker/models/runbot_build.py
```diff
@@ -121,8 +121,10 @@ def job_20_test_all(self, cr, uid, build, lock_path, log_path):
         ] + pr_cmd_env
         logdb = cr.dbname
         if config['db_host'] and not travis_branch.startswith('7.0'):
-            logdb = 'postgres://{cfg[db_user]}:{cfg[db_password]}@' +\
-                    '{cfg[db_host]}/{db}'.format(cfg=config, db=cr.dbname)
+            logdb = 'postgres://%s:%s@%s/%s' % (
```
moylop260, line 124: What about use:

``` python
('postgres://{cfg[db_user]}:{cfg[db_password]}@'
 '{cfg[db_host]}/{db}').format(cfg=config, db=cr.dbname)
```

?

lasley: IMO the string not spanning multiple lines looks better, but I have no explicit preference. I was also adhering to the OCA convention to use `%` over `.format`

moylop260: You are right!

moylop260 (approved):

pedrobaeza (approved):

moylop260:
Thanks!

PR #108, opened 2016-11-20
[ADD] make skipping tests configurable, allow skipping builds too

hbrunn:


runbot_skip_tests/models/runbot_repo.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class RunbotRepo(models.Model):
+    _inherit = 'runbot.repo'
+
+    skip_test = fields.Boolean(
+        'Skip tests', default=True,
+        help='This removes --test-enable from the odoo commandline, '
+        'allowing you to browse a built instance without waiting for tests')
+    skip_build = fields.Boolean(
```
pedrobaeza, line 14: Maybe this option should be in another module?

runbot_skip_tests/models/runbot_build.py
```diff
@@ -0,0 +1,50 @@
...
+        return super(RunbotBuild, self).spawn(
+            cmd,
+            lock_path,
+            log_path,
+            cpu_limit=cpu_limit,
+            shell=shell,
+        )
+
+    def job_10_test_base(self, cr, uid, build, lock_path, log_path):
+        if build.branch_id.repo_id.skip_build:
```
moylop260, line 29: What about use [skip](https://github.com/odoo/odoo-extra/blob/944dd79e004d4e5bce8c5b2d7d8a09e196fb5359/runbot/runbot.py#L910) method from build creation?

hbrunn:
we won't be developing this one further

PR #110, opened 2016-11-27
[ADD] runbot_coverage

hbrunn:
Coverage for runbot
===================

This module was written to have coverage tests in runbot.

Configuration
=============

To configure this module, you need to:

#. check the box `Coverage` on a runbot repository
#. optionally fill in a configuration file
#. optionally fill in the path to you coverage script (``/usr/bin/python-coverage`` for example on debian systems)


runbot_coverage/models/runbot_build.py
```diff
@@ -0,0 +1,88 @@
...
+            ) + cmd[1:]
+        return cmd, modules
+
+    @contextmanager
+    @api.multi
+    def _chdir(self, *args):
+        """Change the current directory, and change back afterwards"""
+        self.ensure_one()
+        current_dir = os.getcwd()
+        os.chdir(self.path(*args))
```
moylop260, line 98: What about use a try-finally to avoid unexpected behaviour?

```python
current_dir = os.getcwd()
os.chdir(self.path(*args))
try:
     yield
except BaseException as berr:
    raise berr
finally:
    os.chdir(current_dir)
```

moylop260: Sorry, the except section is not need it.

```python
current_dir = os.getcwd()
os.chdir(self.path(*args))
try:
     yield
finally:
    os.chdir(current_dir)
```

moylop260: You can see a better example from official documentation of python:
https://docs.python.org/3/library/contextlib.html#contextlib.closing

runbot_coverage/models/runbot_repo.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class RunbotRepo(models.Model):
+    _inherit = 'runbot.repo'
+
+    use_coverage = fields.Boolean('Coverage', default=False)
+    coverage_config = fields.Char(
+        'Configuration file',
+        help='Your coveragerc relative to the build path.')
+    coverage_command = fields.Char(
```
moylop260, line 15: What about use `openerp.tools.which('coverage')` to get this value?
(That method is called to validate the manifest external-python-dependencies)

hbrunn: not a good idea. That's what implicitly happens anyways when you fill in nothing there

runbot_coverage/models/runbot_build.py
hbrunn: done in 7b91045

runbot_coverage/models/runbot_repo.py
moylop260: If is a bad idea use it then, Should we delete the `binary-dependency`?
Because the `odoo.tools.which` is used but without use this field to find the alternative path.
Then odoo won't allow us install it
But if is found and installed then for what we need this additional field?
Could you help me to understand?

hbrunn: yes, the external dependency must go away. The whole dance is necessary for two reasons: On debian, there's no `coverage`, but the thing is called `python-coverage`. So we just can't use the dependencies mechanism. Then second in cases with a weird deployment, you want to use some specific non-system version of the script for isolation reasons. In my case the latter would be pointing to some version of coverage in runbot's home

moylop260 (approved):
Thanks for clarify

moylop260:
@lasley @pedrobaeza @alan196 @dreispt your feedback is welcome

lasley (approved):
Interesting. Code looks good 👍 

I'm curious to see how this will interact with the T2D builds. All my Runbot stuff is v9 & I don't have an easy way to switch, so I'm spitting hypotheticals.

hbrunn:
@lasley I guess not at all. I didn't look yet into the way runbot_docker does it, but probably it'll be better to implement this by including an extra filesystem layer that contains the necessary stuff

susport (commented):
👍

hbrunn:
@moylop260 I count 3 approvals, can we merge this?

alan196 (approved):

Issue #111, opened 2016-12-01
Dockerized Runbot - Missing Image


lasley:
Randomly Runbot is failing with the following message in the all log (from [Runbot build](https://runbot.odoo-community.org/runbot/build/3274021) & https://github.com/OCA/server-tools/pull/611#issuecomment-264187178):

```
Unable to find image 'oca-server-tools:9bb61bf_3' locally
Pulling repository oca-server-tools
time="2016-12-01T14:47:40Z" level=fatal msg="Error: image library/oca-server-tools:9bb61bf_3 not found" 
```

Build looks fine:
```
Sending build context to Docker daemon 24.06 kB

Sending build context to Docker daemon 
Step 0 : FROM vauxoo/odoo-80-image-shippable-auto
latest: Pulling from vauxoo/odoo-80-image-shippable-auto
600f68629688: Already exists
f55680a9f435: Already exists
45aa5b54eb26: Already exists
04e81abeea1d: Already exists
3b8d72e434f8: Already exists
08fe2ef22779: Already exists
45dbfe2621e6: Already exists
fd8638d565b9: Already exists
f9fd3e6c6287: Already exists
e47fd885d98c: Already exists
cb57ee658fe2: Already exists
3d7fb6dd9bff: Already exists
504d62bce149: Already exists
62f906607b33: Already exists
0b0367bd58fd: Already exists
de3c0ca761e0: Already exists
bc1a7c99dee1: Already exists
30a5e802b61e: Already exists
0bf9d24aedcc: Already exists
0f25fe6f3f6c: Already exists
Digest: sha256:73a80d28d41b0c2d5a01951dbe41693a29651a677727c8c9f76db601ee745b4d
Status: Image is up to date for vauxoo/odoo-80-image-shippable-auto:latest
 ---> 0f25fe6f3f6c
Step 1 : ADD files/entrypoint.sh /entrypoint.sh
 ---> 27dedcc9c03a
Removing intermediate container 2713f9a53469
Step 2 : RUN chown -R root:root /entrypoint.sh
 ---> Running in 8d0511a98f0f
 ---> 5016bcedf285
Removing intermediate container 8d0511a98f0f
Step 3 : ADD files/rvm_env.sh /rvm_env.sh
 ---> 04054f378bf0
Removing intermediate container 84ae5c68b6d7
Step 4 : RUN chown -R root:root /rvm_env.sh
 ---> Running in 409969779aa6
 ---> dcb1f4c83687
[...]

PR #114, opened 2016-12-08
[FIX] runbot_travis2docker: Fixing overlapped items

moylop260:
- Helps to fix https://github.com/OCA/runbot-addons/issues/111
- Use a unique name of image
- Use a unique path of t2d scripts
- Remove use of global sys.args (Using directly the t2d class)
- Add _log db and logger in order to self-debug
- Use a copy of ids list instead of original pointer

### DIFF
--- a/runbot_travis2docker/models/runbot_build.py
+++ b/runbot_travis2docker/models/runbot_build.py
@@ -6,7 +6,6 @@
 import logging
 import os
 import time
-import sys
 
 import openerp
 from openerp import fields, models
@@ -16,16 +15,14 @@ from openerp.addons.runbot_build_instructions.models.runbot_build \
 from openerp.addons.runbot.runbot import (
     grep, rfind, run, _re_error, _re_warning)
 
+_logger = logging.getLogger(__name__)
+
 try:
     from travis2docker.git_run import GitRun
-except ImportError:
-    GitRun = None
-try:
-    from travis2docker.cli import main as t2d
-except ImportError:
-    t2d = None
-
-_logger = logging.getLogger(__name__)
+    from travis2docker.cli import get_git_data
+    from travis2docker.travis2docker import Travis2Docker
+except ImportError as err:
+    _logger.debug(err)
 
 
 def custom_build(func):
@@ -61,7 +58,8 @@ class RunbotBuild(models.Model):
     def get_docker_image(self, cr, uid, build, context=None):
         git_obj = GitRun(build.repo_id.name, '')
         image_name = git_obj.owner + '-' + git_obj.repo + ':' + \
-            build.name[:7] + '_' + os.path.basename(build.dockerfile_path)
+            build.name[:7] + '_' + os.path.basename(build.dockerfile_path) + \
+            '_' + str(build.id)
         return image_name.lower()
 
     def get_docker_container(self, cr, uid, build, context=None):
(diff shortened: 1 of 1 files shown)


moylop260:
cc @alan196 @gurneyalex @pedrobaeza @lasley

runbot_travis2docker/models/runbot_build.py
```diff
@@ -61,7 +58,8 @@ class RunbotBuild(models.Model):
     def get_docker_image(self, cr, uid, build, context=None):
         git_obj = GitRun(build.repo_id.name, '')
         image_name = git_obj.owner + '-' + git_obj.repo + ':' + \
-            build.name[:7] + '_' + os.path.basename(build.dockerfile_path)
+            build.name[:7] + '_' + os.path.basename(build.dockerfile_path) + \
```
lasley, line 61: We should just switch to string interpolation here, it's only one more line to the diff and would be nicer IMO

runbot_travis2docker/models/runbot_build.py
```diff
@@ -166,7 +164,7 @@ def job_30_run(self, cr, uid, build, lock_path, log_path):
     @custom_build
     def checkout(self, cr, uid, ids, context=None):
         """Save travis2docker output"""
-        to_be_skipped_ids = ids
+        to_be_skipped_ids = ids[:]
```
lasley, line 167: How is this functionally different than the previous statement?

runbot_travis2docker/models/runbot_build.py
```diff
@@ -175,14 +173,26 @@ def checkout(self, cr, uid, ids, context=None):
...
+                t2d_obj = Travis2Docker(
+                    yml_buffer=yml_content,
+                    work_path=os.path.join(t2d_path, 'script',
+                                           str(build.id) + "_" + sha[:7]),
+                    os_kwargs=git_data,
+                    copy_paths=[("~/.ssh", "$HOME/.ssh")],
+                )
+                path_scripts = t2d_obj.compute_dockerfile(
+                    skip_after_success=True)
+            except BaseException as t2d_e:
```
lasley, line 192: Can you add the to-do into the ReadMe please? I don't like this line - BaseException should never be in an except - so it would be a shame to lose the to-do

lasley (approved):
Minor comments but otherwise 👍

runbot_travis2docker/models/runbot_build.py
moylop260: Use a different pointer with the same values:
```bash
>>> my_list1 = ['a', 'b', 'c']
>>> my_list2 = my_list1
>>> id(my_list2) == id(my_list1)
True
>>> my_list2 = my_list1[:]
>>> id(my_list2) == id(my_list1)
False
```

This avoid change values from original pointer:
```bash
>>> my_list1 = ['a', 'b', 'c']
>>> my_list2 = my_list1
>>> my_list2.remove('a')
>>> my_list1
['b', 'c']  # Delete items from original variable because is a pointer. Bad!

my_list1 = ['a', 'b', 'c']
my_list2 = my_list1[:]
>>> my_list2.remove('a')
>>> my_list1
['a', 'b', 'c']  # Original values is not deleted. Good!
```

runbot_travis2docker/models/runbot_build.py
moylop260: I'm adding a full message in the logger and the db log in order to fix any error from t2d script by each runbot.build.
This is a loop `for` and a error avoid that other builds works.
I don't like BaseException too but IMHO this case is a good exception

runbot_travis2docker/models/runbot_build.py
moylop260: I'm agree with migrate this piece to string interpolation and other similar pieces of code.
But this is unrelated with the focus of this PR.

What about create a issue to work it in other PR?

runbot_travis2docker/models/runbot_build.py
lasley: Ahhh cool nifty trick rather than using copy. Thanks!

gurneyalex (approved):

hbrunn:
and here

PR #118, opened 2017-02-03
[9.0][FIX] runbot_travis2docker: Support ssh URIs

lasley:
* Support URIs prefixed with ssh:// to allow for git over SSH on non-default ports


moylop260 (approved):

pedrobaeza (approved):

lasley:
Woot thanks for the quick-merge - this was breaking my prod CI. I don't want to admit how long it took me to find this 😖

moylop260:
Maybe we should looking for a alternative fix related with [fix repo names and protocol handling with https urls
](https://github.com/gurneyalex/odoo-extra/commit/92b578a4a3bfca8f23dc474357938be7f43fccd6)
@gurneyalex What do you think?

lasley:
Probably a good idea - the fix here still doesn't support `http://` or user prefixes other than `git@`, both of which could still happen.

PR #119, opened 2017-02-22
[ADD] New module for runbot and gitlab integration

JesusZapata:
This module is one alternative to https://github.com/OCA/runbot-addons/tree/10.0/runbot_gitlab module

This module has less code and have only one configuration

Screnshot of configuration 
image

Screenshot of frontend
image


JesusZapata:
@moylop260 
What do you think of this PR?

moylop260 (approved):
This module has minimal changes using the original code and behaviour with inheritance .

> Travis error unrelated because the new version of runbot added a new job `job_21_coverage` and change the expected state for tests.
(I'll fix it from other PR this case)

JesusZapata:
@moylop260 
In this PR now contains the changes made in https://github.com/Vauxoo/runbot-addons/pull/91

runbot_gitlab_minimal/models/runbot_repo.py
```diff
@@ -0,0 +1,209 @@
...
+                    return json
+            except Exception:
+                if ignore_errors:
+                    _logger.exception('Ignored gitlab error %s %r', url,
+                                      payload)
+                else:
+                    raise
+
+
+class RunbotBuild(models.Model):
```
moylop260: Use a file by each class
`runbot_build.py` and `runbot_repo.py`

JesusZapata: @moylop260
Thanks for you comment

moylop260:
@hbrunn your feedback is welcome

moylop260:
@tremlin  @PCatinean your feedback is welcome

hbrunn:
@moylop260 yes, reviewing this is on my list for a while. Can't do a proper review right now, but I think cleaner code never hurts.
It does seem to contain some idiosyncratic code like https://github.com/OCA/runbot-addons/pull/119/files#diff-851bd8f22c1876a4aaa87c8f61ea98e2R20, but I'll need more time for properly looking into this.

In any case, I think you should replace my version of the gitlab integration instead of adding a second one, and make it very override friendly. Then I'd add the features missing from 8.0 in a module on top of this instead of having duplicated code that basically does the same thing.

moylop260:
@hbrunn Thanks for feedback


@JesusZapata 
Could you identify a PR branch using the `ref name` if starts with `pull/` like as [original-runbot-module](https://github.com/odoo/odoo-extra/blob/851320c0175ce21082eb39272005a80e028de4c5/runbot/runbot.py#L502):

jesus

JesusZapata:
@hbrunn
Thanks for you comment, Change the way to calculate if the branch is merged request

JesusZapata:
@moylop260 
Now on this PR have partial change made on https://github.com/Vauxoo/runbot-addons/pull/98

JesusZapata:
@moylop260 
Now this PR contains this change https://github.com/Vauxoo/runbot-addons/pull/111

runbot_gitlab/models/runbot_build.py
```diff
@@ -0,0 +1,79 @@
...
+
+from .runbot_repo import _get_url, _get_session
+
+_logger = logging.getLogger(__name__)
+
+
+class RunbotBuild(models.Model):
+    _inherit = "runbot.build"
+
+    def github_status(self, cr, uid, ids, context=None):
```
lasley, line 18: Why not use the new API?

runbot_gitlab/models/runbot_build.py
```diff
@@ -0,0 +1,79 @@
...
+        runbot_domain = self.pool['runbot.repo'].domain(cr, uid)
+        for build in self.browse(cr, uid, ids, context=context):
+            is_merge_request = build.branch_id.name.startswith('refs/pull')
+            source_project_id = False
+            _url = _get_url('/projects/:owner/:repo/statuses/%s' % build.name,
+                            build.repo_id.base)
+            if not build.repo_id.uses_gitlab:
```
lasley, line 25: This should go immediately after the for - it renders the lines above useless otherwise

runbot_gitlab/models/runbot_build.py
```diff
@@ -0,0 +1,79 @@
...
+                super(RunbotBuild, self).github_status(cr, uid, ids,
+                                                       context=context)
+                continue
+            if not build.repo_id.token:
+                continue
+            session = _get_session(build.repo_id.token)
+            try:
```
lasley, line 32: This try/except is rather broad. Is it possible to trim it down to smaller, explicit controls?

runbot_gitlab/models/runbot_build.py
```diff
@@ -0,0 +1,79 @@
...
+                if is_merge_request:
```
lasley, line 33:
```
if not is_merge_request:
    continue
```
to reduce nesting of positive case

runbot_gitlab/models/runbot_build.py
```diff
@@ -0,0 +1,79 @@
...
+                    response.raise_for_status()
+                    json = response.json()[0]
+                    source_project_id = json['source_project_id']
+                    if source_project_id:
+                        url = _get_url('/projects/%s' % source_project_id,
+                                       build.repo_id.base)
+                        response = session.get(url)
+                        response.raise_for_status()
+                        json = response.json()
+                        base_url = (json['web_url'].replace('http://', '').
```
lasley, line 47: IMO you should `urlparse` for this

```
parsed_url = urllib.parse.urlparse(json['web_url'])
base_url = parsed_url.netloc
```

runbot_gitlab/models/runbot_build.py
```diff
@@ -0,0 +1,79 @@
...
+                                    replace('https://', ''))
+                        _url = _get_url('/projects/:owner/:repo/statuses/%s' %
+                                        build.name, base_url)
+                desc = "runbot build %s" % (build.dest,)
+                if build.state == 'testing':
+                    state = 'running'
+                elif build.state in ('running', 'done'):
+                    state = 'failed'
+                    if build.result == 'ok':
```
lasley, line 56: IMO this should be a dictionary of mappings that are either returned from a method or defined in the instance. This would cleanup the if/then chain, as well as allow for inherited changes.

runbot_gitlab/models/runbot_build.py
```diff
@@ -0,0 +1,79 @@
...
+                    if build.result == 'skipped':
+                        state = 'canceled'
+                    if build.result == 'killed':
+                        state = 'canceled'
+                    desc += " (runtime %ss)" % (build.job_time,)
+                else:
+                    continue
+                status = {
+                    "state": state,
+                    "target_url": "http://%s/runbot/build/%s" % (runbot_domain,
```
lasley, line 69: Hardcoding http?

IMO you should use the `parsed_url` above; `scheme` property.

runbot_gitlab/models/runbot_repo.py
```diff
@@ -0,0 +1,154 @@
...
+
+import requests
+
+from openerp import models, fields
+
+_logger = logging.getLogger(__name__)
+
+
+def _get_url(url, base):
+    match_object = re.search('([^/]+)/([^/]+)/([^/.]+(.git)?)', base)
```
lasley, line 19: I'm honestly lost on this method. Would it be possible to maybe use named groups? I think that might help the readability.

lasley (commented):
Thanks! We've been using this in prod for a few months now. Works great. Few comments

lasley:
Also - it would be nice to see tests here.

hbrunn:
@JesusZapata are you going to follow up on @lasley's comments? If not, I think I'll take your commits and port them to 10, do you agree?

JesusZapata:
@hbrunn 
Yes, I agree

hbrunn:
great, I'll do that in the coming weeks then I guess.

General question: Are you all still betting on runbot given the odoo.sh thing, or are you looking to migrate away from this as CI/CD platform? If so, where to? jenkins? gitlab?

hbrunn:
should we close this for the 11 version or are you going to keep developing here?

moylop260:
Superseded by https://github.com/OCA/runbot-addons/pull/151

PR #121, opened 2017-03-23
[IMP] runbot_travis2docker: Security and ssh features

moylop260:
- Use a non-sudoer user
  - Updating travis2docker use `odoo` user by default
- Start ssh service with `root`
- Add ssh keys for github committer and author

NOTE: This change requires a upgrade of travis2docker `pip install -U travis2docker`

For merger: Use github button `Merge and squash`

### DIFF
--- a/runbot_travis2docker/__openerp__.py
+++ b/runbot_travis2docker/__openerp__.py
@@ -6,7 +6,7 @@
 {
     "name": "Runbot travis to docker",
     "summary": "Generate docker with odoo instance based on .travis.yml",
-    "version": "9.0.1.1.0",
+    "version": "9.0.1.2.0",
     "category": "runbot",
     "website": "https://odoo-community.org/",
     "author": "Vauxoo,Odoo Community Association (OCA)",
--- a/runbot_travis2docker/migrations/9.0.1.2.0/post-migrate.py
+++ b/runbot_travis2docker/migrations/9.0.1.2.0/post-migrate.py
@@ -1,0 +1,10 @@
+# coding: utf-8
+# © 2017 Vauxoo
+#   Coded by: moylop260@vauxoo.com
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version):
+    """Skip new feature for old runbot builds."""
+    cr.execute("UPDATE runbot_build SET docker_executed_commands = true")
+    cr.commit()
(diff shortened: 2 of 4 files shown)


lasley (commented):
Dude. Yes. 

High level code looks good. I'm installing this on my Runbot in the next few hours. Will report back soon & perform a more thorough review

yajo:
If I understand it well, this is installing and running an sshd server in each runbot job, right? What's that for?

moylop260:
Hello @Yajo thanks for comment.
Debug from ssh is like as circle feature: https://circleci.com/docs/1.0/ssh-build

> Often the best way to troubleshoot problems is to ssh into a running or finished build to look at log files, running processes, and so on.
You can do that with CircleCI!

...and Runbot too!

There are many OCA issues related with `locally we can't reproduce the travis or runbot error`
IMHO CircleCI have a good point because the best environment to reproduce it is the self-environment.

OCA PRs had errors related with a particular version of: phantomjs, postgresql, lxml... or a missed/extra local configuration.
But is not so easy to identify it if we can't reproduce it.

Note: The current image is using a new user named `odoo` (not sudoer)

lasley:
I fully agree that I could have solved the recent Phantom issue in significantly less time had I been able to SSH into the build. 

I had to send something like 5 different builds in order to get something as simple as the PhantomJS version to print out correctly in the log. At 10-15 minutes a build, that means I spent about an hour to obtain the Phantom version. Admittedly I was doing other things too, so it's not like I was just twiddling my thumbs waiting, but still would have been nice to be able to complete in one fell swoop.

moylop260:
@lasley Thanks for comment.
We have a similar case for postgresql version from:
https://github.com/OCA/maintainer-quality-tools/issues/432#issuecomment-281580935

yajo (approved):
I don't really understand what this PR does, but I like it.

lasley (approved):
Code + functional, excellent feature @moylop260

moylop260:
@lasley 
thank for test it

@pedrobaeza 
Could we merge it?

gurneyalex:
I'll deploy this today.

moylop260:
Thanks I just tested it

I'm connecting using the instance port plus one for build
 - https://runbot.odoo-community.org/runbot/repo/github-com-oca-server-tools-149

borrar

Using:
`ssh -p $((8718 + 1)) odoo@runbot1.odoo-community.org`

The result was successfully:
screen shot 2017-03-26 at 2 00 43 pm

Issue #122, opened 2017-03-30
[IMP] runbot: too slow first time connection generating routing map


moylop260:
Steps to reproduce it
===============

 - We have a [build runbot instance](https://runbot.odoo-community.org/runbot/build/3284913) with the following [running log](https://runbot1.odoo-community.org/runbot/static/build/3284913-10-0-d09a11/logs/job_30_run.txt)
```txt
...
2017-03-29 17:29:56,473 167 INFO openerp_test odoo.modules.loading: Modules loaded.
...
```

Opening first time the url of the instance and second one using `time python -c "import urllib;urllib.urlopen('http://3284913-10-0-d09a11.runbot1.odoo-community.org/?db=3284913-10-0-d09a11-all')"` the result is:
![screen shot 2017-03-29 at 6 30 26 pm](https://cloud.githubusercontent.com/assets/6644187/24482210/d2e24790-14ac-11e7-8966-09bae31b2130.png)

The log affected was:
```txt
...
2017-03-30 00:21:40,762 167 INFO ? odoo.http: HTTP Configuring static files
2017-03-30 00:21:40,768 167 INFO openerp_test odoo.addons.base.ir.ir_http: Generating routing map
2017-03-30 00:21:43,674 167 INFO openerp_test werkzeug: 172.17.42.1 - - [30/Mar/2017 00:21:43] "GET /?db=3284913-10-0-d09a11-all HTTP/1.0" 200 -
2017-03-30 00:21:47,414 167 INFO openerp_test werkzeug: 172.17.42.1 - - [30/Mar/2017 00:21:47] "GET /?db=3284913-10-0-d09a11-all HTTP/1.0" 200 -
```

Saw
===
Log output  time spent 3 seconds `from 00:21:43 to 00:21:47`
Opening url time spent 4.5 seconds first time
Opening url time spent 0.8 seconds second time

Expected
=======
Open url time 4.5 seconds for first time connection for real user.

I mean, we could add a open url simulation from runbot after up build instance from runbot to improvement usability.

moylop260:
Closing since that it is an issue opened a long time ago without a good welcome

PR #123, opened 2017-04-04
[REF] runbot_travis2docker: Faster

moylop260:
- [x] Fix https://github.com/OCA/runbot-addons/issues/122
  - [x] Open url after finish build in order to simulate the-first-user to generate routing map and static files
- [x] Remove `debug=1` because is too slow and is not used for all cases.
- [x] Remove database default name based on `build.dest` field because now is used just one database `openerp_test` (default name from MQT)

Note: It requires update [runbot master branch](https://github.com/odoo/odoo-extra/tree/master/runbot) to get the new original method [runbot._get_branch_quickconnect_url](https://github.com/odoo/odoo-extra/blob/851320c0175ce21082eb39272005a80e028de4c5/runbot/runbot.py#L474)

### DIFF
--- a/runbot_travis2docker/models/__init__.py
+++ b/runbot_travis2docker/models/__init__.py
@@ -5,3 +5,4 @@
 
 from . import runbot_build
 from . import runbot_repo
+from . import runbot_branch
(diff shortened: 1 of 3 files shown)


hbrunn:
also here, still relevant?

PR #130, opened 2017-06-12
8.0 runbot buildout

hbrunn:
Buildout for runbot
===================

This module was written to make it convenient to use [buildout based build](http://pythonhosted.org/anybox.recipe.odoo>) on runbot.

Configuration
=============

To configure this module, you need to:

- turn on `Use buildouts` on a repository
- fill in a regex that allows to differentiate a normal branch name: For runbot, branches should be named ``$version-$sometext`` anyways, so this module's suggestion is to call buildout branches ``buildout-$version-$sometext``
- check the value of the field `Buildout section`, this must be the name you use in your buildouts
- if you use multiple buildouts for the same version to implement some kind of DTAP scenario (you should), you can mark some buildout branch as the one to be used for tests by navigating to the branch and checking `Default for this version`
- when converting a repository to use buildouts, be sure to remove all dependency branches first. Otherwise, runbot will copy all of them unnecessarily. Then, rebuild some buildout branch, and when it is green, rebuild another branch that is supposed to use the buildout

Background
==========

This module will cause buildout branches to be treated radically differently: For buildout branches, the buildout will be run, and any failures reported in the build results. For normal branches, it will search for a buildout branch with a matching version, copy this buildout to a temporary directory, manipulate buildout.cfg to use the branch to be tested and the ports assigned for this instance, then rerun the buildout, then run tests within the buildout environment.

For this to work, your buildout.cfg needs to contain one addons line::

    git $the_exact_repo_url_as_configured_in_runbot parts/$repo_name $version

or if you use pinned branches::

    git $the_exact_repo_url_as_configured_in_runbot parts/$repo_name $commit_hash branch=$version

Which then will be replaced by::

    git $the_exact_repo_url_as_configured_in_runbot parts/$repo_name $commit_hash branch=$current_branchname


hbrunn:
superseded by #154

PR #131, opened 2017-06-14
[FIX] runbot_travis2docker: Adding support to base languages

JesusZapata:
This PR adding support to base languages
For each language available get the prefix and group by this

Example: `es_ES > es`

image

Related with https://github.com/Vauxoo/runbot-addons/pull/130

### DIFF
--- a/runbot_travis2docker/models/runbot_repo.py
+++ b/runbot_travis2docker/models/runbot_repo.py
@@ -24,6 +24,8 @@ class RunbotRepo(models.Model):
     @api.constrains('weblate_languages')
     def _check_weblate_languages(self):
         supported_langs = [item[0] for item in scan_languages()]
+        supported_langs.extend(set([lang.split('_')[0] for lang in
+                                    supported_langs]))
         for record in self.filtered('weblate_languages'):
             langs = record.weblate_languages.split(',')
             for lang in langs:


JesusZapata:
@moylop260 
In this PR contains the change made in https://github.com/Vauxoo/runbot-addons/pull/130

moylop260 (approved):

pedrobaeza:
I don't get it how it works. Can you tell me?

JesusZapata:
@pedrobaeza 
I previously only supported languages https://github.com/odoo/odoo/blob/9.0/openerp/tools/misc.py#L481 the files `es.po, fr.po, pt.po, en.po ...` are not supported

This change for each language supported in odoo obtains the base name of the language
For example : `es_AR > es (Spanish), fr_CA > fr (French), pt_PT > pt (Portuguese), en_US > en (English) ...`

pedrobaeza:
OK, but this can cause incorrect available languages. If you only have `pt_BR` translations for example, then it will report `pt` as available.

JesusZapata:
@pedrobaeza 
If in the view only select `pt_BR` that will be the only language supported in weblate
This change is referring to the constrains, now support base languages `es, pt, en, ...`

pedrobaeza:
OK, I see. Thanks for the explanations.

pedrobaeza (approved):

PR #134, opened 2017-07-28
[REF] runbot_travis2docker: Adding new enrironment variable WEBLATE_SSH

JesusZapata:
This variable should be used to made the fetch of the weblate git repository

This change is necessary because weblate does not work well when trying to put security to the internal repository

Previously the weblate gitexport module was used to download the changes from weblate https://docs.weblate.org/en/latest/vcs.html#https-repositories now weblate should be enable ssh access to the vcs folder

Related with https://github.com/Vauxoo/runbot-addons/pull/133


JesusZapata:
@moylop260 
In this PR has the same solution of https://github.com/Vauxoo/runbot-addons/pull/133

PR #136, opened 2017-08-17
[9.0][IMP] runbot_travis2docker: Add custom image

lasley:
* Add ability to define custom image for travis2docker builds
* Default to standard `vauxoo/odoo-80-shippable` to make the image more obvious to user

### DIFF
--- a/runbot_travis2docker/models/runbot_repo.py
+++ b/runbot_travis2docker/models/runbot_repo.py
@@ -15,10 +15,17 @@ class RunbotRepo(models.Model):
 
     is_travis2docker_build = fields.Boolean('Travis to docker build')
     travis2docker_test_disable = fields.Boolean('Test Disable?')
+    travis2docker_image = fields.Char(
+        default=lambda s: s._default_travis2docker_image(),
+    )
     weblate_url = fields.Char(default="https://weblate.odoo-community.org/api")
     weblate_token = fields.Char()
     weblate_languages = fields.Char(help="List of code iso of languages E.g."
                                     " en_US,es_ES")
+
+    @api.model
+    def _default_travis2docker_image(self):
+        return 'vauxoo/odoo-80-image-shippable-auto'
 
     @api.multi
     @api.constrains('weblate_languages')
--- a/runbot_travis2docker/README.rst
+++ b/runbot_travis2docker/README.rst
@@ -65,6 +65,7 @@ Contributors
 ------------
 
 * Moisés López <moylop260@vauxoo.com>
+* Dave Lasley <dave@laslabs.com>
 
 Maintainer
 ----------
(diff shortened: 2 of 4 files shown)


moylop260 (approved):

runbot_travis2docker/models/runbot_repo.py
```diff
@@ -15,11 +15,18 @@ class RunbotRepo(models.Model):
...
+        default=lambda s: s._default_travis2docker_image(),
+    )
     weblate_url = fields.Char(default="https://weblate.odoo-community.org/api")
     weblate_token = fields.Char()
     weblate_languages = fields.Char(help="List of code iso of languages E.g."
                                     " en_US,es_ES")
 
+    @api.model
+    def _default_travis2docker_image(self):
+        return 'vauxoo/odoo-80-image-shippable-auto'
```
elicoidal, line 28: We should avoid company specific information in the code.

elicoidal (changes requested):

runbot_travis2docker/models/runbot_repo.py
pedrobaeza: But that's the current image.

pedrobaeza (approved):

runbot_travis2docker/models/runbot_repo.py
lasley: Yeah right now this is obscured deep in [Travis2Docker](https://github.com/Vauxoo/travis2docker/blob/0622bd064cd277791bb9997858130799869331ff/src/travis2docker/travis2docker.py#L69), so I'm just bringing it up a layer.

moylop260: What about skip the parameter if the value is Null in order to avoid use a default value from here?

lasley: I set the field to required, which circumvents the possibility of a null. I felt that falling back to the core default obscures the image too much.

Admittedly half of the reason I did this is that I have forgotten which image T2D runs on at least 10 occasions & had to dig into the code. We're pretty far into layers here, so I felt it would be best to explicitly expose that layer.

I'll change if you guys feel otherwise, I just wanted to explain this was deliberate and elaborate on my reasoning.

lasley: Also worth noting the other half of this ambition is to provide an OCA supported build image. More to come on that soon.

moylop260: @elicoidal Are you agree? Can we merge it?

elicoidal (approved):

runbot_travis2docker/models/runbot_repo.py
elicoidal: I understand and wont block for that  but over time we might need to think about an OCA repo in docker

PR #142, opened 2017-09-29
[REF] travis_requirements: Download big image

moylop260:
Currently CI is red because a big docker image is download during test running and expire the max time waiting


moylop260:
Merging since that change from red to green status build result

pedrobaeza:
Yes, but you're hardcoding again a resource outside OCA (from vauxoo). We should remove that dependencies ASAP.

moylop260:
Yeah, you are right. It was discussed here: https://github.com/OCA/runbot-addons/pull/136#discussion_r134000099

PR #145, opened 2017-12-29
[REF] runbot_travis2docker: Commit into the weblate repository if the repository have a changes uncommitted

JesusZapata:
In some cases, the user that makes on Weblate not has permission to make a commit over the Weblate repository, if the commit is not done the process never run

For example, when the user makes translated and this change is not committed the Weblate repository does not have this change 

image

When this error happens the only way to run the Weblate integration is using the button `Force Weblate`

Related to the following pull request https://github.com/Vauxoo/runbot-addons/pull/145

### DIFF
--- a/runbot_travis2docker/models/runbot_branch.py
+++ b/runbot_travis2docker/models/runbot_branch.py
@@ -94,6 +94,15 @@ class RunbotBranch(models.Model):
                     break
                 page += 1
             project['components'] = components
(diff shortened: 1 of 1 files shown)


JesusZapata:
@moylop260 
This is only one cherry-pick of the following pull request https://github.com/Vauxoo/runbot-addons/pull/145

PR #132, opened 2017-07-24
[IMP] runbot_travis2docker: Move run cmd to helper method

lasley:
* Move the docker run command generation to a helper method to allow for better inherited usage

I'm making a module to allow for the use of a Traefik load balancer in a Dockerized environment, and require access to the run args in order to append the proper labels.

PR #147, opened 2018-01-19
[FIX] travis_requirements: Allow freeze a compatible version of runbot module

moylop260:
Fix https://github.com/OCA/runbot-addons/pull/132#issuecomment-357814884

NOTE: Here we can to do so much but now we need a small fix in order to have travis green to merge other PRs


lasley (approved):
Thanks @moylop260

PR #154, opened 2018-04-26
[ADD] runbot_buildout

hbrunn:


runbot_buildout/models/runbot_build.py
```diff
@@ -0,0 +1,335 @@
...
+
+    @api.multi
+    def _server(self, *l, **kw):
+        self.ensure_one()
+        if self.repo_id.uses_buildout:
+            odoo_part = 'parts/%s' % self.repo_id.buildout_section
+            if os.path.exists(self._path(odoo_part, 'odoo')):
+                return self._path(odoo_part, 'odoo', *l, **kw)
+            return self._path(odoo_part, 'openerp', *l, **kw)
+        else:
```
moylop260: `else-return` is not needed if you `if` section uses a `return`
You can use just `return ...` without `else`

moylop260: What about use return early?

```python
if not  self.repo_id.uses_buildout:
    return ...
# My code for uses_buildout without nested `if` and without `else` reduce cyclomatic complex
```

runbot_buildout/models/runbot_build.py
```diff
@@ -0,0 +1,335 @@
...
+            build._log('buildout', 'Running buildout')
+            return build._spawn_buildout(
+                [
+                    build._path('bin/buildout'), '-N', '-q',
+                    '-c', buildout_file,
+                ],
+                lock_path, log_path,
+            )
+        else:
+            return super(RunbotBuild, self)._job_00_init(
```
moylop260: if I have `build.repo_id.uses_buildout=True` but don't have `build.branch_id.buildout_version` should continue the normal runbot process?

runbot_buildout/models/runbot_build.py
```diff
@@ -0,0 +1,335 @@
...
+            build.write({'result': 'ko'})
+            result = False
+        if not result:
+            build._github_status()
+        return result
+
+    @api.multi
+    def _spawn_buildout(self, cmd, lock_path, log_path):
+        self.ensure_one()
+        out = open(log_path, "w")
```
moylop260: When is this file closed?

runbot_buildout/models/runbot_build.py
```diff
@@ -0,0 +1,335 @@
...
+            ], stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as exception:
+            self._log('buildout', exception.output)
+            self.write({
+                'state': 'done',
+                'result': 'ko',
+            })
+            self._github_status()
+            return False
+        buildout_file = tempfile.NamedTemporaryFile(
```
moylop260: What about use `with`?
```python
with tempfile... as buildout_file:
    buildout_file.write(adapted_buildout)
```

runbot_buildout/tests/test_runbot_buildout.py
```diff
@@ -0,0 +1,22 @@
...
+        branch = self.env['runbot.branch'].create({
+            'name': 'refs/heads/buildout-11.0-testing',
+            'repo_id': repo.id,
+        })
+        self.assertTrue(branch.buildout_version)
+        build = self.env['runbot.build'].create({
+            'branch_id': branch.id,
+            'name': '42',
+        })
+        self.assertIn('parts/', build._server())
```
moylop260: Is possible add a real tests running all jobs using a repo of example real?

moylop260:
Oh sorry I didn't the WIP label

hbrunn:
thanks for your input anyways, some of your points I didn't think about

runbot_buildout/models/runbot_build.py
hbrunn: by now a few lines below, thanks for this one

hbrunn:
@moylop260 this is reviewable by now. Seems like there was an incompatible change in runbot: https://travis-ci.org/OCA/runbot-addons/jobs/386271662#L2709

moylop260:
@hbrunn 
FYI I have fixed from
https://github.com/OCA/runbot-addons/pull/160

runbot_buildout/static/src/js/runbot_buildout.js
```diff
@@ -0,0 +1,16 @@
+//-*- coding: utf-8 -*-
```
daramousk: Minor change, but do we actually need this?

runbot_buildout/models/runbot_build.py
```diff
@@ -0,0 +1,434 @@
...
+            self._github_status()
+            return False
+
+        with tempfile.NamedTemporaryFile(
+                prefix='buildout', suffix='.cfg', dir=self._path(),
+                delete=False,
+        ) as buildout_file:
+            buildout_file.write(bytes(
+                '[buildout]\n'
+                'extensions = buildout.wheel\n'
```
thomaspaulb: Probably we then also have to change this?

runbot_buildout/static/src/js/runbot_buildout.js
hbrunn: not at all, thanks. That's a fluke from my addons template, fixed it in the template and will remove this on next push

runbot_buildout/models/runbot_build.py
hbrunn: yes, and I'm going to ruhtlessly steal your snippet to use newer setuptools

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: python
 
 python:
-  - "3.5"
+  - "3.6"
```
moylop260: Why python 3.6?
I mean, odoo uses by default 3.5

.travis.yml
```diff
@@ -25,16 +25,21 @@ env:
...
-  - TESTS="1" ODOO_REPO="odoo/odoo"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="runbot_buildout"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="runbot_travis2docker"
 
 
 install:
+  # only install extra dependencies when we're not linting
+  - pip install zc.recipe.egg --no-cache-dir
+  # the latest package on pip doesn't speak python3
+  - pip install https://github.com/anybox/anybox.recipe.odoo/tarball/master/anybox.recipe.odoo-1.9.3.dev0.tar.gz
```
moylop260: Is not good use requirements.txt for this case, right?

runbot_buildout/models/runbot_build.py
moylop260: Ok, What about use `with open(log_path, "w")` in order to ensure that this file is closed if there is an error in the middle.

.travis.yml
hbrunn: i have no idea. Must have been a leftover from me working frantically to get the tests running, there were very weird errors

.travis.yml
hbrunn: I tried to pin this to a git+https url, but that didn't work the first couple of tries and then I figured I'll just wait it out until this version is released

moylop260 (approved):
👍

moylop260:
Do you have planned squash the commits or you prefer that I use merge&squash github button?

hbrunn:
I squashed them, thanks!

runbot_buildout/models/runbot_build.py
```diff
@@ -0,0 +1,454 @@
...
+    @api.model
+    def _job_00_init(self, build, lock_path, log_path):
+        if build.repo_id.uses_buildout and\
+                not build.branch_id.buildout_version:
+            # this is a normal code branch, find a working buildout
+            # for it and build it
+            build._github_status()
+            buildout_build = build._get_buildout_build()
+            if not buildout_build:
+                return -2
```
daramousk: Maybe it would be better if those return codes where defined as class variables with self-explanatory names.

runbot_buildout/tests/test_runbot_buildout.py
```diff
@@ -0,0 +1,89 @@
...
+        git('commit', '-m', 'initial commit')
+
+        repo = self.env['runbot.repo'].create({
+            'name': git_dir,
+            'uses_buildout': True,
+        })
+
+        repo._update(repo)
+        repo._scheduler(repo.ids)
+
```
daramousk: Could the code from here to the beginning of the function be moved in `setUp` so that we can clean this up a little bit?

runbot_buildout/tests/test_runbot_buildout.py
```diff
@@ -0,0 +1,89 @@
...
+        self.assertEqual(code_build.job, 'job_01_run_buildout')
+        code_build._schedule()
+        os.waitpid(code_build.pid, 0)
+        self.assertIn('parts/', code_build._server())
+        self.assertTrue(os.path.exists(code_build._server()))
+        cmd, modules = code_build._cmd()
+        self.assertTrue(any('start_odoo' in c for c in cmd))
+        self.assertFalse(modules)
+        builds._kill()
+        repo.unlink()
```
daramousk: Besides stuff being moved into `setUp` maybe this can be refactored a little bit more so that we can have clear understanding of what is tested:

In this function the tests are on:
1) job_10_test_base
2) job_20_test_all
3) job_00_init
4) job_01_run_buildout

Maybe each test should have its own `test_$job_name` and also a `tearDown` function at the end that does all the cleaning.
https://docs.python.org/2/library/unittest.html#unittest.TestCase.tearDown

hbrunn:
@daramousk @thomaspaulb done

daramousk (approved):

.travis.yml
```diff
@@ -25,19 +25,23 @@ env:
...
-  - TESTS="1" ODOO_REPO="odoo/odoo"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="runbot_buildout"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="runbot_travis2docker"
 
 before_install:
   # Fix https://github.com/travis-ci/travis-ci/issues/8982#issuecomment-354357640
   - python3.5 -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
 
 install:
+  - pip install zc.recipe.egg --no-cache-dir
```
moylop260: Did you try using requirements.txt file?

hbrunn: i'll try now

PR #155, opened 2018-05-10
[DEL] runbot_subject_skip: Not installable

moylop260:
Because this behavior is now supported from runbot officially

From https://github.com/odoo/runbot/pull/13


moylop260:
Question:
 - What is the best way to deprecate it?
   - Remove it
   - Uninstallable False

elicoidal (approved):
LGTM (but why not simply remove the module?)

pedrobaeza:
But AFAIK, we are using a runbot fork, isn't it? Is our fork updated?

moylop260:
I didn't get the fork:
 - https://github.com/oca/runbot is not exists.

Are you talking about https://runbot.odoo-community.org/runbot?

We are not have migrated to 11.0 version yet.
That runbot is using odoo 9.0 yet and this change just affects 11.0.

pedrobaeza:
Ah, OK

pedrobaeza (approved):

moylop260:
The question here is
Should we delete the module?

pedrobaeza:
Yes, better to remove it in this branch.

PR #157, opened 2018-05-29
[FIX] runbot_travis2docker: Remove db parameter in URLs when t2d is used

luisg123v:
When builds are generated using travis2docker, the database name will
always be set to `openerp_test`. Therefore, the db name given by default
by Runbot should not be used in the build' sign in URL.

This change ensures URLs include the db name parameter only when the
build was not generated using travis2docker.

In addition, this hides the "Connect base" button when the build was not
generated using travis2docker, because the database *-base is not
created in such cases.


luisg123v:
CC @moylop260

moylop260 (approved):

PR #158, opened 2018-05-29
[IMP] requirements.txt: Update travis2docker version

luisg123v:
This is to support the key `postgresql`, to define the PostgreSQL
version.

For more info, see:
https://github.com/Vauxoo/travis2docker/pull/117

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 matplotlib==2.2.2
-travis2docker==4.0.2
+travis2docker==4.0.7


luisg123v:
CC @moylop260

requirements.txt
```diff
@@ -1,2 +1,2 @@
 matplotlib==2.2.2
-travis2docker==4.0.2
+travis2docker==4.0.6
```
moylop260: Use 4.0.7 instead.
I have created a bumpversion now with that patch.

luisg123v: Done

moylop260 (approved):
This change adds support for the key `addons` -> `postgresql`, located
in the file `.travis.yml`. That key is used to manually specify the
PostgreSQL version, instead of using the default one.

Currently we are using a global environment variable for runbot in the .travis.yml
plus a addons->postgresql for travis in the .travis.yml

Then 👍

PR #159, opened 2018-06-04
[ADD] runbot_preseed_database

hbrunn:
This module allows you to have runbot use a database template that already has some modules installed. This allows for much faster builds in case you're only interested in a subset of the tests.


sbidoul:
@hbrunn talking about pre-seeding databases, you may be interested in [click-odoo-initdb](https://github.com/acsone/click-odoo-contrib#click-odoo-initdb-beta).

hbrunn:
good product placement ;-) but seriously, this looks very nice. I'd like to recycle the runbot machinery here however, so that a database is actually created by a build (and in my case, this build is done by `runbot_buildout`)

PR #164, opened 2018-06-21
[ADD] map runbot's `warn` result to gitlab's `success_with_warnings`

hbrunn:
@moylop260 so this was my naive approach to implement the warnings. Didn't work, and I was convinced to do something stupid, until I looked this up at the gitlab side:
https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/api/commit_statuses.rb#L98

So next step is to check if it's just as simple as adding the warning status there, but my ruby is not really existing. Any pointers are appreciated.

### DIFF
--- a/runbot_gitlab/models/runbot_build.py
+++ b/runbot_gitlab/models/runbot_build.py
@@ -49,6 +49,8 @@ class RunbotBuild(models.Model):
                     state = 'failed'
                     if build.result == 'ok':
                         state = 'success'
+                    if build.result == 'warn':
+                        state = 'success_with_warnings'
                     if build.result == 'ko':
                         state = 'failed'
                     if build.result == 'skipped':


moylop260:
Did you mean create a new PR to gitlab-ci in order to support this case?

hbrunn:
yes. but first i have to get familiar with ruby in order to provide a meaningful mr there. that's then more a pet project of mine and will stay in 'work in progress' quite a while. or i will propose to make warning succeed with an amendment that there were warnings. that would be great for my workflow with colleagues - what do you think?

moylop260:
Yes, if the CIs support a middle status like as runbot (yellow result) is a meaningful for us too.

FYI I requested the same for travis-ci:
 - https://github.com/travis-ci/travis-ci/issues/4730

hbrunn:
I fear the gitlab people won't be much more forthcoming... For my own purposes, succeeding with a different name or different description already would suffice, how about just doing this instead?

hbrunn:
superseded by #180 as I don't think the api will be extended any time soon

Issue #162, opened 2018-06-18
Weblate from runbot_travis2docker is used for new weblate release?


moylop260:
Currently weblate is based from .pot
Then too many code was deprecated.
The question is if we should deprecate the code here too

https://github.com/OCA/runbot-addons/blob/9.0/runbot_travis2docker/models/runbot_repo.py#L34

cc @sbidoul @benwillig

sbidoul:
Yes, runbot is not used for the weblate integration. This code can be removed too.

moylop260:
Fixed from https://github.com/OCA/runbot-addons/pull/166

PR #166, opened 2018-06-27
[11.0][REM] rem all references to weblate

umiphos:
In this PR we are purging weblate, because runbot is not used for the weblate integration in 11.0 version. 
related to #162


moylop260 (approved):
LGTM
👍 

Could you bump version of the module, please?

hbrunn (changes requested):
you need to remove the files you delete from the manifest too to fix https://travis-ci.org/OCA/runbot-addons/jobs/397449834#L1099

umiphos:
@hbrunn shame on me, it's now fixed, thanks

pedrobaeza (changes requested):
You have to change the module version number and include a migration script that removes the cron (if exists).

hbrunn:
come on pedro. this is not customer stuff, this is our own tools. i think we can expect users here to fix stuff on their own

hbrunn (approved):

pedrobaeza:
OK, it's not so difficult, but I don't insist

pedrobaeza (approved):

PR #174, opened 2018-08-28
[REF] .travis.yml: Fix stdout error

moylop260:
Random message when run 'docker pull IMAGE' like
 - [Errno 11] write could not complete without blocking

Seems as a travis issue:
 - https://github.com/travis-ci/travis-ci/issues/8982#issuecomment-354357640
 - https://github.com/travis-ci/travis-ci/issues/4704#issuecomment-348435959

Currently stable branch has travis red:
 - https://travis-ci.org/OCA/runbot-addons/builds/407457537

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,9 @@ env:
   - TESTS="1" ODOO_REPO="OCA/OCB"
   - TESTS="1" ODOO_REPO="odoo/odoo"
 
+before_install:
+  # Fix https://github.com/travis-ci/travis-ci/issues/8982#issuecomment-354357640
+  - python3.5 -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
 
 install:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I have found this error in some builds, but isn't a better solution? Should we put this workaround in all the repositories?

hbrunn (approved):

hbrunn:
I'm not aware of a better fix, and would propose to apply it only where it hurts

hbrunn:
@moylop260 doesn't seem to have helped :-( https://travis-ci.org/OCA/runbot-addons/jobs/428502161#L1090

moylop260:
Ok Thank you
The strange of this is that is not posible reproduce it from a PR.
It just is reproduced it from a stable branch.
What about create a new issue for travis-ci?
-- 
Moisés López Calderón
Mobile: (+521) 477-752-22-30
Twitter: @moylop260
hangout: moylop260@vauxoo.com
http://www.vauxoo.com - Odoo Gold Partner
Twitter: @vauxoo

PR #178, opened 2018-09-07
[FIX] runbot_gitlab: Retrieve correctly SSH keys when under GitLab CI

luisg123v:
Currently, SSH keys are not retrieved correctly, due to a bad parsing of
the HTTP response. When keys are retrieved from GitLab, the response's
content is not a json, but a raw page; so it's raw bites, not str.

This commit parses correctly such responses, so SSH keys may be parsed
correctly and SSH access may be granted.

This also causes empty authors to be skipped, they of course have
neither username nor SSH keys, and retrieving them from GitLab would
cause issues.


moylop260 (approved):
👍 
I just checked that is working fine the feature for both github and gitlab

pedrobaeza (approved):

Issue #176, opened 2018-09-04
travis2docker runs: postgresql is not found


gurneyalex:
I see a lot of builds not available to interact with. In the job_30_xxx.txt logs of these we get

```
======== Testing test_server.py ========
command:  /entrypoint_image
 * Starting PostgreSQL 9.6 database server                                       * Removed stale pid file.
The PostgreSQL server failed to start. Please check the log output:
2018-09-04 11:24:13 UTC [125]: [1-1] db=,user=FATAL:  lock file "/var/run/postgresql/.s.PGSQL.5432.lock" already exists
2018-09-04 11:24:13 UTC [125]: [2-1] db=,user=HINT:  Is another postmaster (PID 103) using socket file "/var/run/postgresql/.s.PGSQL.5432"?
2018-09-04 11:24:13 UTC [125]: [3-1] db=,user=LOG:  database system is shut down
                                                                         [fail]
Waiting to start psql service...
Traceback (most recent call last):
  File "/entrypoint_image", line 120, in <module>
    docker_entrypoint()
  File "/entrypoint_image", line 98, in docker_entrypoint
    raise RuntimeError("PSQL not started.")
RuntimeError: PSQL not started.
Working in /home/odoo/build/OCA/account-invoicing
Using repo OCA/OCB and addons path /home/odoo/build/OCA/account-invoicing,/home/odoo/OCB-11.0/addons
Modules to test: ['account_invoice_check_total', 'account_invoice_force_number', 'account_invoice_line_description', 'account_invoice_refund_link', 'account_invoice_reimbursable', 'account_invoice_supplier_ref_reuse', 'account_invoice_supplier_ref_unique', 'account_invoice_supplier_self_invoice', 'account_invoice_tax_required', 'account_invoice_triple_discount', 'account_invoice_view_payment', 'account_payment_term_extension', 'sale_timesheet_invoice_description']
Modules to preinstall: ['account', 'decimal_precision', 'sales_team', 'account_invoicing', 'project', 'sale_management', 'base_setup', 'web_planner', 'sale_timesheet', 'web_tour', 'base', 'sale', 'http_routing', 'resource', 'web', 'product', 'bus', 'mail', 'analytic', 'portal', 'hr', 'hr_timesheet']

Creating instance:
createdb: could not connect to database template1: could not connect to server: Connection refused
        Is the server running locally and accepting
        connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
Using previous openerp_template database.

[...]

PR #179, opened 2018-09-10
[FIX] runbot_travis2docker: Stop postgresql to avoid corruption

moylop260:
Fix https://github.com/OCA/runbot-addons/issues/176

Merger: Squash and merge my pr please

### DIFF
--- a/runbot_travis2docker/models/runbot_build.py
+++ b/runbot_travis2docker/models/runbot_build.py
@@ -144,6 +144,9 @@ class RunbotBuild(models.Model):
                 branch_short_name, '--root-path=' + t2d_path,
                 '--exclude-after-success',
                 '--docker-image=%s' % build.repo_id.travis2docker_image,
+                # Avoid corruption of postgresql
+                '--runs-at-the-end-script=pg_isready -q && '
+                '/etc/init.d/postgresql stop'
             ]
             try:
                 path_scripts = t2d()
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 matplotlib==2.2.2
-travis2docker==4.0.7
+travis2docker==4.1.0


pedrobaeza (approved):

runbot_travis2docker/models/runbot_build.py
```diff
@@ -144,6 +144,10 @@ def _checkout(self):
                 branch_short_name, '--root-path=' + t2d_path,
                 '--exclude-after-success',
                 '--docker-image=%s' % build.repo_id.travis2docker_image,
+                # Avoid corruption of postgresql
+                '--runs-at-the-end-script=pg_isready -q && '
+                '/etc/init.d/postgresql stop && '
+                'rm -f /var/run/postgresql/.s.PGSQL.5432*',
```
luisg123v: Why do you need to remove this file? Shouldn't it be automatically removed if everything goes OK when stopping the PostgreSQL service?

In case this is really needed, I'd like to suggest adding the `-v` parameter, to be able to log and know when this file was manually removed:  
`rm -fv /var/run/postgresql/.s.PGSQL.5432*`

moylop260: This is a recommendation of @gurneyalex from 
https://github.com/Vauxoo/travis2docker/pull/119

I think it is uneeded but he tested this part and it is working.
Using verbose mode lets to see in the log what cases is removing it manually

moylop260: @luisg123v 
The new build is not showing the deleting of the file.
Then I will remove this part of the code.

https://travis-ci.org/OCA/runbot-addons/jobs/426694697#L1594

luisg123v (approved):

PCatinean (approved):

PCatinean:
Tested on a running instance, all builds go through. Thank you for your contribution LGTM :+1:

moylop260:
@pedrobaeza
Could we merge it?

PR #180, opened 2018-09-14
[IMP] pass builds with warnings as success to gitlab

hbrunn:
let's show builds with warnings as success but add a comment that there's a warning

### DIFF
--- a/runbot_gitlab/models/runbot_build.py
+++ b/runbot_gitlab/models/runbot_build.py
@@ -49,6 +49,12 @@ class RunbotBuild(models.Model):
                     state = 'failed'
                     if build.result == 'ok':
                         state = 'success'
+                    if build.result == 'warn' and\
+                            not build.repo_id.warnings_are_errors:
+                        # as long as the gitlab api doesn't support
+                        # success_with_warnings, we pass success and a comment
+                        state = 'success'
+                        desc = "WARNING\n" + desc
                     if build.result == 'ko':
                         state = 'failed'
                     if build.result == 'skipped':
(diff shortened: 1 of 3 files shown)


moylop260:
Should be this one a github behavior too?

Is not this change hidding warnings, since that the people don’t open the CI to check the final result?
currently, travis has warnings for this reason.

But lets to hear other opinions

@pedrobaeza @gurneyalex @rvalyi

pedrobaeza:
I prefer to have any kind of highlighting. Runbot warnings are important (it's not the same as pylint ones).

hbrunn:
of course warnings are important, and as soon as we can show them in gitlab somehow, let's please do that. This is a temporary proposal until the API allows something like #164

pedrobaeza:
OK, as you think it's the best. We are not using runbot anymore with our doodba-qa (https://github.com/Tecnativa/doodba-qa) + Gitlab CI solution.

hbrunn:
I think I'll add this as an option. We'll want to treat warnings as errors often, but there are projects where we're happy to have at least no failures for the time being. And as the project matures, you can turn this on again.

ap-wtioit:
I support this pull request (we already do similar stuff on our runbot), especially the solution with the context enables the warning to be  visible in gitlab as well (we don't have that yet)

moylop260:
+1 for me as an option.

hbrunn:
now as an option

moylop260 (approved):

pedrobaeza (approved):

PR #181, opened 2018-09-18
[FIX] runbot_gitlab: repair sending coverage data to gitlab

ap-wtioit:
runbot_gitlab was not correctly reporting coverage to gitlab, this is fixed with this commit

`.coverage` is boolean and marks if coverage was enabled in build.
`.coverage_result` contains the result in percent as gitlab expects it


hbrunn (approved):

ap-wtioit:
Image with the result in gitlab (3 builds without coverage, 1 build with new coverage data): https://support.wt-io-it.at/owncloud/index.php/s/JQvkpG5RdlDi6lw

PR #184, opened 2018-10-19
[REF] runbot_gitlab: Change the way GitHub icons and texts are replaced

luisg123v:
Instead of e.g.:
Inherit a GitHub icon and add a condition to replace the class by the GitLab's one in case it's a GitLab repo, or leave untouched otherwise.

Now the following is performed:
- Hide elements for GitLab repos
- Add a new element that will be visible  only for GitLab repos

This new approach makes easier to support new backends in the future (e.g. Bitbucket), because they won't cause conflicts, they would only need to follow the same approach.

In addition, not only icons are taken into account, but also the text "Open commit on GitHub".

### DIFF
--- a/runbot_gitlab/templates/build.xml
+++ b/runbot_gitlab/templates/build.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <template id="build_button_inherit" inherit_id="runbot.build_button">
+        <!-- Replace icon and change title: Open commit on GitHub -> GitLab -->
+        <xpath expr="//a[@aria-label='Open commit on GitHub']" position="attributes">
+            <attribute name="t-att-invisible" add="(repo.uses_gitlab and 1)" separator=" or "/>
+        </xpath>
+        <xpath expr="//a[@aria-label='Open commit on GitHub']" position="after">
+            <a t-att-invisible="not repo.uses_gitlab and 1"
+                t-attf-href="https://#{repo.base}/commit/#{bu['name']}"
+                class="btn btn-default"
+                title="Open commit on GitLab"
+                aria-label="Open commit on GitLab">
+                <i class="fa fa-gitlab"/>
+            </a>
+        </xpath>
+    </template>
+
+</odoo>
--- a/runbot_gitlab/__manifest__.py
+++ b/runbot_gitlab/__manifest__.py
@@ -12,6 +12,7 @@
     "depends": ["runbot"],
     "data": [
         "views/runbot_repo.xml",
+        "templates/build.xml",
         "templates/runbot_gitlab_logos.xml",
     ],
     "installable": True,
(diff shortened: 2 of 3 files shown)


moylop260 (commented):
Is posible use a dynamic value?
Since that we could have bitbucket in the future or another one... and it could be hard to manage it

luisg123v:
@moylop260  
I'd like to do so, but I'd say it's not possible. This is because it was not designed that way, the type of repository should be a selection field, instead of a boolean `is_gitlab`, `is_bitbucket`, etc.

I agree with your point, but I think that, given the current model structure, this is the best approach we may take.

moylop260 (approved):

runbot_gitlab/templates/build.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <template id="build_button_inherit" inherit_id="runbot.build_button">
+        <!-- Change title: Open commit on GitHub -> GitLab -->
+        <xpath expr="//a[@aria-label='Open commit on GitHub']" position="attributes">
```
pedrobaeza, line 6: For being more resilient, I would do:

- Add on current buttons an invisible condition to be hidden when they are not GitHub.
- Add new buttons with the new labels for Gitlab
- If you have a new backend like Bitbucket, it should add the same

luisg123v: @pedrobaeza  
I like the idea, it sounds more scalable. However, that would need to be a separate work, because it would imply a little of refactoring

That button I'm inheriting, along with other ones, are already inherited and modified by this same module, e.g.:

https://github.com/OCA/runbot-addons/blob/62a839f8ab8f3af2994790f4fa27d149317fd493/runbot_gitlab/templates/runbot_gitlab_logos.xml#L14-L17

Actually, that's why I took this approach, to be more consistent with what it's already done.

pedrobaeza: Then switch the other template to the same technique.

hbrunn:
@luisg123v will you followup on @pedrobaeza's proposal?

luisg123v:
@hbrunn  
Yes. However, I'd like, if possible, to do it on a separate PR, because I think that would be more of a refactor that a fix.

@pedrobaeza  would you agree on approving this and leave the refactor for another PR?

pedrobaeza:
Sorry but no. This is not so big for doing it in 2 things, as this smells like you will delay a lot the second task.

luisg123v:
@pedrobaeza  
I agree on your point that the second task will be delayed, but currently both tasks are being delayed because of the second one.

On the oter hand, I don't agree on your point that the change is not so big, because it will require to modify all inheritances, and IMO should be a separate change.

But well, if you don't agree, I'll try to perform those changes this weekend.

runbot_gitlab/templates/build.xml
luisg123v: @pedrobaeza & @moylop260 
I applied the proposed approach. Could you review, please?

Sorry for the delay, I've been so busy these days. But I'm in the game again.

pedrobaeza (approved):

luisg123v:
@moylop260  
Could you merge, please?

PR #185, opened 2018-12-10
[FIX] runbot_gitlab: to prevent exceptions with runbot addon from 1330d5209879cebc

ap-wtioit:
we construct a head label for gitlab as well, otherwise generating the
branch_url will throw an exception in _get_branch_url()
https://github.com/odoo/runbot/commit/1330d5209879cebc7351c5730374ddda23221678#diff-f845984e99fe832a8c029aa2d2cc8d3aR44

Info @wt-io-it


moylop260 (approved):

moylop260:
Fast tracking since that it is compatible with both versions of runbot addon and it is adding a new key supported by gitlab.

PR #186, opened 2018-12-21
11.0 fix runbot gitlab branch infos

ap-wtioit:
After some investigation on our runbot / gitlab i discovered (and fixed) some more issues in runbot_gitlab that happen after https://github.com/odoo/runbot/commit/1330d5209879cebc7351c5730374ddda23221678 introfuced the target_branch_name field

* target_branch_name is set to the actual source branch
* getting information for multiple pull requests at once only returned the information of the last pull request for all of the pull requests

Info @wt-io-it

### DIFF
--- a/runbot_gitlab/models/runbot_repo.py
+++ b/runbot_gitlab/models/runbot_repo.py
@@ -101,7 +101,7 @@ class RunbotRepo(models.Model):
                         else response.text)
                 if 'merge_requests/' in url:
                     json['head'] = {
-                        'ref': json['target_branch'],
+                        'ref': json['source_branch'],
                         # github api returns a label like
                         # 'source-project-name:source-branch-name' the closest
                         # we got in gitlab api is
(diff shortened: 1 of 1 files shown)


hbrunn:
it's hard for me to understand what's going on here. Can you elaborate?

ap-wtioit:
@hbrunn we had the problem that the gitlab api is not correctly implemented, /merge_requests?iid=x does not select the merge request with iid x but /merge_requests/x does (https://docs.gitlab.com/ce/api/merge_requests.html#list-merge-requests vs https://docs.gitlab.com/ce/api/merge_requests.html#get-single-mr). Therefore getting the details for a merge request did get you details of all the merge requests in the given project, the first one in the list was then picked, which was always the newest merge request in the project (the one with the highest iid). This works fine as long as there is only one merge request opened at a time, but it breaks if you open two or more of them between the tow runbot _checkout calls.

When i debugged this problem i also noticed that source and target branch are mixed up in comparison to the github api. So i also fixed that (This is used to detect the correct upstream branch of the merge request. And the upstream branch is then used to select the correct branch of the repositories dependencies for the build). Also there is a pull request pending in odoo runbot that fixes the name matching there: https://github.com/odoo/runbot/pull/62

More detail should also be available in the two commit messages:
https://github.com/OCA/runbot-addons/pull/186/commits/4201cd54d3c2283ae45920d325e28524c5530b3a
https://github.com/OCA/runbot-addons/pull/186/commits/8dd190e60b1dc1511255063064bf90f2a827be15

hbrunn:
thanks, this might be also the reason why our gitlab never shows builds as green when they've been red before. Code looks good, will do some tests locally

hbrunn (approved):

PR #188, opened 2019-01-16
[ADD] pin runbot to the last non-docker version

hbrunn:
with https://github.com/odoo/runbot/commit/4c0cd91914425a4c73494397b6f3d17c46fd0627, runbot changed drastically and we'll have to adapt all of our codebase to this.

I propose that for the time being, we pin the dependency to before docker, and for version 12, use master again.

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -1,1 +1,1 @@
-runbot https://github.com/odoo/runbot master
+runbot https://github.com/odoo/runbot pre-docker


hbrunn:
PS: I noticed we don't support pins in MTQ, so I used a temporary fork. As that's not optimal, I asked in https://github.com/odoo/runbot/issues/76 if we can have a tag on the main repo for that

hbrunn:
PPS: cool, the tag is there

moylop260 (approved):
👍 
It PR is sticking the stable version used for this repository.

pedrobaeza (approved):

PR #189, opened 2019-01-16
[FIX] don't pass status from old builds for the same commit

hbrunn:
this finally gives us greenness in gitlab when rebuilding a failed build


hbrunn:
actually, that's a bug in runbot fixed in 4f1a55da9beb1099a03c02034a8900d63571f4b2

PR #190, opened 2019-01-23
[11.0][FIX] runbot buildout correct interreter

NL66278:
Prevent Odoo 11 (and higher) branches from being run by python2.

### DIFF
--- a/runbot_buildout/models/runbot_branch.py
+++ b/runbot_buildout/models/runbot_branch.py
@@ -1,7 +1,11 @@
-# Copyright 2017 Therp BV <http://therp.nl>
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# Copyright 2017-2019 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 import re
 from openerp import api, fields, models
+
+
+PYTHON2 = '/usr/bin/python2'
+PYTHON3 = '/usr/bin/python3'
 
 
 class RunbotBranch(models.Model):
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):
Seems OK. Code review

runbot_buildout/models/runbot_branch.py
```diff
@@ -35,3 +39,21 @@ def _compute_buildout_version(self):
             if not match:
                 continue
             this.update({'buildout_version': match.groupdict()['version']})
+
+    @api.multi
+    def get_interpreter(self):
```
moylop260, line 44: What about get interpreter from magic comment of odoo-bin file?

hbrunn: that doesn't exist, repos that rely on buildouts never have any plain odoo/ocb repo in runbot, just the buildout and the code/module branches

runbot_buildout/models/runbot_build.py
```diff
@@ -414,7 +415,7 @@ def _adapt_buildout(self):
                 'options.without_demo = False\n'
                 'options.data_dir = %(datadir)s\n'
                 'options.dbfilter = %(dbfilter)s\n'
-                'options.lang = en_US' % dict(
+                'options.load_language = en_US' % dict(
```
hbrunn, line 418: see https://github.com/OCA/runbot-addons/pull/187, but keep it here, then I'll close mine

runbot_buildout/models/runbot_branch.py
NL66278: @moylop260 we need the interpreter to bootstrap the buildout. This bootstrap wil create the bin/buildout script that, after running the buildout, will create the odoo part that contains odoo-bin.

runbot_buildout/models/runbot_build.py
NL66278: @hbrunn It got mixed up/in because my starting point was your branch. Now merged anyway.

PR #194, opened 2019-09-03
[FIX] have explicit modules on a branch override repo modules, not add to them

hbrunn:
no idea why I though they should be merged before

### DIFF
--- a/runbot_buildout/models/runbot_build.py
+++ b/runbot_buildout/models/runbot_build.py
@@ -144,8 +144,7 @@ class RunbotBuild(models.Model):
                     if '/%s/' % parts_path in dirname:
                         repo_modules.append(basename)
             explicit_modules = (
-                (build.branch_id.modules or '') + ',' +
-                (build.repo_id.modules or '')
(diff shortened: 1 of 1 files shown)


runbot_buildout/models/runbot_build.py
```diff
@@ -144,7 +144,7 @@ def _job_10_test_base(self, build, lock_path, log_path):
                     if '/%s/' % parts_path in dirname:
                         repo_modules.append(basename)
             explicit_modules = (
-                (build.branch_id.modules or '') + ',' +
+                (build.branch_id.modules or '') or
```
moylop260: What about `build.branch_id.modules or build.repo_id.modules or ''`?

moylop260 (approved):
small nitpick but not blocking

👍

runbot_buildout/models/runbot_build.py
hbrunn: much better, thanks

moylop260:
Red CI is unrelated for module changed here.
Merging since that it is small fix and there are not too much reviewers for this module.

PR #197, opened 2020-01-29
[ADD] runbot_push_pot

hbrunn:
with this, we can have runbot export pot files for all modules being tested, and push them back to our repository. This way, you can have a completely self hosted gitlab runbot gitlab weblate loop.


runbot_push_pot/static/src/js/runbot_push_translation.js
```diff
@@ -0,0 +1,6 @@
+//Copyright 2019 Therp BV <https://therp.nl>
```
NL66278: I think this file and the empty css file can be removed.

hbrunn: agreed. I took a note to look into the CI of this repo, this is horribly complex and fails often, as now.

NL66278 (approved):
:+1: LGTM

hbrunn:
failures seem to be caused by this running with python 3.5 instead of 3.7

runbot_push_pot/models/runbot_build.py
```diff
@@ -0,0 +1,86 @@
...
+
+                git('checkout', branch_name)
+                translations.extractall(checkout)
+                git('add', '.')
+                diff_lines = [
+                    line for line in
+                    subprocess.run(
+                        ['git', '-C', checkout, 'diff', 'HEAD', '-U0'],
+                        capture_output=True
+                    ).stdout.decode('utf8').split('\n')
```
moylop260, line 74: What about using dual compatibility methods?
I mean,
```suggestion
                    subprocess.check_output(
                        ['git', '-C', checkout, 'diff', 'HEAD', '-U0'],
                    ).decode('utf8').split('\n')
```

PR #199, opened 2020-05-14
[ADD][runbot_buildout] improvements

hbrunn:
- don't have a flag `buildout_default` on builds, but set a default buildout on the repo. The original was meant to support multiple versions in the same repo, but this turns out awkward and usually you only temporarily have two versions. This you can deal with with branch specific buildout branches then.
- don't git repack -a, this might drop objects we need later
- more logging
- mute sentry for all buildouts

### DIFF
--- a/runbot_buildout/migrations/11.0.1.1.0/post-migration.py
+++ b/runbot_buildout/migrations/11.0.1.1.0/post-migration.py
@@ -1,0 +1,11 @@
+# Copyright 2020 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version=None):
+    cr.execute(
+        """update runbot_repo set buildout_branch_id=runbot_branch.id
+        from runbot_branch where
+        runbot_branch.repo_id=runbot_repo.id and runbot_branch.buildout_default
+        """
+    )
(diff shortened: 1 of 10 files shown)


runbot_buildout/migrations/11.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
```
NL66278: No coding here ;-)

hbrunn: thanks. I'm stilly busy making tests work on runbot, so this PR might get a little bit noisy

hbrunn:
I believe the complaints from codepatch are a fluke, it shows at least some file from runbot_docker as the one not being tested enough, and this isn't in the diff here...

after merging this, I think we can rebase https://github.com/OCA/runbot-addons/pull/197 and then this should come back green too

NL66278 (approved):
:1st_place_medal: LGTM

thomaspaulb (approved):
LGTM

moylop260:
it can't be merged by ocabot since that there is a unrelated error then merging manually

hbrunn:
thanks @moylop260 

Question: Therp uses the pre-docker branch of runbot in v11 because she never did the step to container stuff. Are you still busy with this in v11, or are you using the v13 runbot and do reviews here as a courtesy (in that case, thank you so much)?

moylop260:
We are using pre-docker too.

